PR #10, opened 2015-12-15
test(utils): Add tests for getHighlightedValue

pixelastic:
Given a hit returned by the API and a key, will return the highlighted value if one is present, or the default key otherwise. Added some tests to it, and moved it to utils.


redox:
LGTM :+1:

PR #12, opened 2015-12-15
test(utils): Add tests for getSnippettedValue

pixelastic:
Same underlying logic than getHighlightedValue, but also add prefix
and suffix ellipsis (…) when we detect that a snippet is truncating
a sentence.


redox:
Go!

PR #13, opened 2015-12-15
chore(release): add release process

vvo:
+ add specific algolia agent to the js client
+ add npm run release, npm run release:beta
+ add CONTRIBUTING guide
+ finish rename to /docsearch

### DIFF
--- a/scripts/build-css
+++ b/scripts/build-css
@@ -2,9 +2,8 @@
 
 set -e # exit when error
 
-VERSION=$(json version < package.json)
-NAME='docsearch.js'
-LICENSE="/*! ${NAME} ${VERSION:-UNRELEASED} | © Algolia | github.com/algolia/docsearch.js */"
+NAME='docsearch'
+LICENSE="/*! ${NAME} ${VERSION:-UNRELEASED} | © Algolia | github.com/algolia/docsearch */"
 DIST_DIR="dist/cdn"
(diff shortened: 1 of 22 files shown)


vvo:
After this one I will do a test release to check the whole process is ok

scripts/build-css
```diff
@@ -2,9 +2,8 @@
 
 set -e # exit when error
 
-VERSION=$(json version < package.json)
-NAME='docsearch.js'
-LICENSE="/*! ${NAME} ${VERSION:-UNRELEASED} | © Algolia | github.com/algolia/docsearch.js */"
+NAME='docsearch'
```
pixelastic: Why did you remove the `VERSION`?

vvo: the VERSION is passed by the release script to the build script.

pixelastic: Can we still manually run `npm run build:css|js` or will it fail?

vvo: It will work, the version will be "UNRELEASED" in the licence file that's it.

When would you need to do so? (Or just curious?)

pixelastic: The example in `dev` uses symlinks that points to the css files in `dist`, so I need to run `npm run build:css` when I run `npm run dev`

vvo: ok it will work yes

PR #18, opened 2015-12-16
Add branding footer

redox:


src/lib/DocSearch.js
```diff
@@ -43,7 +43,8 @@ class DocSearch {
     this.autocomplete = autocomplete(this.input, autocompleteOptions, [{
       source: this.getSource(),
       templates: {
-        suggestion: this.getSuggestionTemplate()
+        suggestion: this.getSuggestionTemplate(),
+        footer: '<div class="ads-footer">Search by <a class="ads-footer--logo" href="https://www.algolia.com/docsearch">Algolia</a></div>'
```
pixelastic: Can't we make the whole footer a link instead? Larger area to click.

redox: Actually that was on purpose, but maybe it could make sense.

pixelastic: So, do you think the little link is enough?

redox: Hmmm, yes?

PR #19, opened 2015-12-16
Lighter default colors

redox:
screen shot 2015-12-16 at 19 14 41


pixelastic:
Oh yes, much better.

redox:
OK, lets start with that then!

PR #22, opened 2015-12-18
test(getAutocompleteSource): Add tests for source

pixelastic:
I had to change some methods to static to be able to tests them
without instantiating a new DocSearch. I also used better `restore`
methods of sinon.js.

I did not manage to correctly test the promise returned by the method
returned by getAutocompleteSource. I would have like to test that when
resolved it correctly call reformatHits on the data as well as call
the callback on the reformated hits but got confused in
spying/stubbing/mocking the promise.


redox:
:green_heart:

PR #35, opened 2015-12-21
fix(css): Renaming all `ads-` prefix to `algolia-docsearch-`

### LINKED ISSUE #34: Rename CSS classes to not use `ads-` prefix
`ads` stands for Algolia DocSearch, but it is blocked by ad blockers (tested with µBlock). We will have to rename our classes to use something else.

pixelastic:
Fixes #34


redox:
:+1:

PR #38, opened 2015-12-22
docs(readme): Add info about styling and local dev

pixelastic:
I've added a screencast in the readme, a screenshot of the default
colorscheme and a copy-paste of the CSS customization example.

The diff of the README is all red because I used
[formd](http://drbunsen.github.io/formd/) on the file to make the
links more readable (it simply uses reference instead of inlining).


redox:
Awesome :100:

PR #42, opened 2015-12-23
chore(css): ensure the menu is wide enough, but no !important here

redox:


pixelastic:
Did you check on small mobile devices?

redox:
Arg no

redox:
@pixelastic safer like that

src/styles/main.scss
```diff
@@ -111,6 +109,9 @@
 // BREAKPOINT 1:
 // Screen is big enough to display the text snippets
 @media (min-width: $breakpoint-medium) {
+  .aa-dropdown-menu {
+    min-width: 400px;
```
pixelastic: Could you use the `$breakpoint` variables instead of hardcoding the width?

redox: Hmmm, something like `$breakpoint-medium * 0.75` ?

pixelastic: Yes maybe, or just putting 400px in a variable defined at the top so it feels a bit less magic (and overridable by users).

redox: Go for the variable then.

PR #60, opened 2015-12-28
fix(url): Removes concatenation of URL with hash if it contains a hash

### LINKED ISSUE #53: Remove the url+anchor concatenation
There is something confusing with the way we store the `url` and `anchor`.

The scraper index both a `anchor` field and a `url` one, but the `url` already contains the anchor. We do that because we're doing a `distinct` on the `url`. But the JS client also concatenate the `url` and `anchor`, which results in weird things.

I think we should:
- Remove the `anchor` field from the scraper
- Remove the concatenation from the JS client

bobylito:
FIX #53

### DIFF
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -130,7 +130,7 @@ class DocSearch {
 
     // Translate hits into smaller objects to be send to the template
     return groupedHits.map((hit) => {
-      let url = hit.anchor ? `${hit.url}#${hit.anchor}` : hit.url;
+      let url = DocSearch.formatURL(hit);
       let category = utils.getHighlightedValue(hit, 'lvl0');
       let subcategory = utils.getHighlightedValue(hit, 'lvl1') || category;
       let displayTitle = utils.compact([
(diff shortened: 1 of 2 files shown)


redox:
:+1: LGTM

src/lib/DocSearch.js
```diff
@@ -154,6 +154,21 @@ class DocSearch {
...
 
+  static formatURL(hit) {
+    const {url, anchor} = hit;
+    if (url) {
+      const containsAnchor = url.indexOf('#') !== -1;
+      if (containsAnchor) return url;
+      else if (anchor) return `${hit.url}#${hit.anchor}`;
+      return url;
+    }
+    else if (anchor) return `#${hit.anchor}`;
```
pixelastic: When would that happen? If we end up with a record with no `url` but an `anchor` field, then we have an issue with the crawler :)

test/DocSearch-test.js
```diff
@@ -708,6 +708,70 @@ describe('DocSearch', () => {
...
+          lvl4: null,
+          lvl5: null
+        },
+        content: 'foo bar',
+        url: 'http://foo.bar/#anchor',
+        anchor: 'anchor'
+      }];
+
+      // When
+      let actual = DocSearch.formatHits(input);
```
pixelastic: I think we should test `formatURL` directly, not as a side effect of `formatHits`.

PR #68, opened 2016-01-27
add docsearch

maxiloc:
image


pixelastic:
It's failing because of the naming convention on the commit messages. We're following the "conventional changelog" of Angular (like for instantsearch).

Just rename it to something like "docs(docsearch): Adding DocSearch to DocSearch" ;)

maxiloc:
@pixelastic Fixed

PR #71, opened 2016-01-29
Dropdown usability improvement

seafoox:
In the current version, each row in the left column (display sub categories) are separated with a line, even when two suggestions belong to the same sub-category. 

screenshot 2016-01-29 16 16 21

Removing those, and making the interlines lighter in the right column, improves readability.
Makes it easier for the brain to understand that some items belong to a same sub-category.

screenshot 2016-01-29 16 15 40

Also disabled the mouse hover look'n'feel when hovering the left column.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -21,8 +21,12 @@
   color: #333;
   cursor: pointer;
   overflow: hidden;
-  border-bottom: 1px solid $color-border;
 }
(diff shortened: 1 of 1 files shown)


seafoox:
Re-submitted the pull request (see https://github.com/algolia/docsearch/pull/72) to follow the commit convention

PR #72, opened 2016-01-29
fix(dropdown): Visually group suggestions together

seafoox:
In the current version, each row in the left column (display sub categories) are separated with a line, even when two suggestions belong to the same sub-category. 

screenshot 2016-01-29 16 16 21

Removing those, and making the interlines lighter in the right column, improves readability.
Makes it easier for the brain to understand that some items belong to a same sub-category.

screenshot 2016-01-29 16 15 40

Also disabled the mouse hover look'n'feel when hovering the left column.

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -21,8 +21,12 @@
   color: #333;
   cursor: pointer;
   overflow: hidden;
-  border-bottom: 1px solid $color-border;
 }
+
+  // Each sub-suggestion
+ .algolia-docsearch-suggestion__secondary {
+    border-top: 1px solid $color-border;
+  }
 
 // Main category headers
 .algolia-docsearch-suggestion--category-header {
@@ -62,6 +66,7 @@
 // The secondary column is hidden on small screens
 .algolia-docsearch-suggestion--subcategory-column {
   display: none;
+  cursor: default;
 }
 // The text snippet is hidden on small screens
 .algolia-docsearch-suggestion--text {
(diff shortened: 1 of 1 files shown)


src/styles/main.scss
```diff
@@ -62,6 +66,7 @@
 // The secondary column is hidden on small screens
 .algolia-docsearch-suggestion--subcategory-column {
   display: none;
+  cursor: default;
```
Jerskouille: Not sure it's such a good idea to remove the `cursor: pointer` on something clickable.

Jerskouille:
:100:

src/styles/main.scss
seafoox: Tried to disable the click using `pointer-events: none;` but didn't work. I guess we'd need to disable  in the JS by checking if the parent of the click elements if different from the left column.

For now when a user click it redirect to the page of suggestion that is at the same level as the mouse. And not to the sub-category, like the user could expect.

pixelastic:
Currently, clicks on the dropdown are intercepted by the js and "manually" redirect to the destination. This means that `ctrl+click` does not work. We'll have to fix this. I think @Jerskouille worked on something similar already?

Jerskouille:
I haven't.
However, in `instantsearch.js` we solved that with links, we could try something like:
- have links in the markup to handle clicks + control clicks
- have the `autocomplete:selected`.event return if it's a click, and redirect as before if it's with the keyboard
I'm not sure how to detect it though.

PR #73, opened 2016-02-08
Fixed markdown links inside code blocks

redox:
Ensure the links are correctly interpreted by the documentation website.

screen shot 2016-02-08 at 01 41 08


Jerskouille:
I think the correct conventional changelog wording is `docs`.

vvo:
(Both are ok because we made the mistake often) :D

PR #79, opened 2016-03-09
fix(algoliaOptions): ensure we keep default options

### LINKED ISSUE #78: Default hitsPerPage not applied if algoliaOptions are specified
As soon as you specify some `algoliaOptions`, we lose the default `hitsPerPage` value. Would be better to keep that value as a default.

redox:
While overriding the algoliaOptions, we were losing the default
hitsPerPage query parameters.

Fix #78

### DIFF
--- a/test/DocSearch-test.js
+++ b/test/DocSearch-test.js
@@ -111,6 +111,36 @@ describe('DocSearch', () => {
       // Then
       expect(actual.appId).toEqual('foo');
     });
+    it('should allow customize algoliaOptions without loosing default options', () => {
+      // Given
+      let options = {
+        algoliaOptions: {
+          facetFilters: ['version:1.0']
+        },
+        ...defaultOptions
+      };
+
+      // When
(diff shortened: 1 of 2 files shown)


test/DocSearch-test.js
```diff
@@ -136,7 +166,7 @@ describe('DocSearch', () => {
       let actual = new DocSearch(options);
 
       // Then
-      expect(actual.algoliaOptions).toEqual('algoliaOptions');
+      expect(actual.algoliaOptions).toEqual({hitsPerPage: 5});
```
pixelastic: The whole PR LGTM except for this part. This test seems unrelated. I would rather set an option of `{hitsPerPage: 10}` and test that.

We want to test that it accepts the options we pass, while this test assert that we keep the default value.

redox: Right, fixing now :+1:

redox:
@pixelastic updated :+1:

PR #86, opened 2016-03-24
Consider the key's lowercase form while grouping

### LINKED ISSUE #85: utils.groupBy should be case-insensitive
On some websites, the categories on which we're grouping on don't share the same case. For the grouping, it makes a lot of sense to actually consider the lowercase form only.

redox:
Fix #85

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -35,6 +35,8 @@ let utils = {
   },
   /*
   * Group all objects of a collection by the value of the specified attribute
+  * If the attribute is a string, use the lowercase form.
+  *
   * eg.
   * groupBy([
   *   {name: 'Tim', category: 'dev'},
@@ -69,6 +71,9 @@ let utils = {
         throw new Error(`[groupBy]: Object has no key ${property}`);
       }
       let key = item[property];
+      if (typeof key === 'string') {
+        key = key.toLowerCase();
+      }
       if (!newCollection[key]) {
         newCollection[key] = [];
       }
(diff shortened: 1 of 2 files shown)


redox:
@pixelastic What do you think about that?

src/lib/utils.js
```diff
@@ -68,7 +68,7 @@ let utils = {
       if (item[property] === undefined) {
         throw new Error(`[groupBy]: Object has no key ${property}`);
       }
-      let key = item[property];
+      let key = item[property] && item[property].toLowerCase();
```
pixelastic: Why the `&&`? 

`let key = item[property].toLowerCase();` should be enough

maxiloc: I think playground was failing otherwise

redox: Because `item[property]` can be `null`

pixelastic: Then we should update the `if` a few line above, where it is testing if it's `undefined` to also test `null`

redox: Hmmm, I think it's fine to group hits sharing the `null` value together. It happens if - for a unknown reason - you don't have `lvl1` for instance.

I can make the test more explicit if you want.

pixelastic: I would make the code more explicit then.

```
let key = item[property];
if (key.toLowerCase) {
  let = key.toLowerCase();
}
```

Still not perfect (`key` can be an object with a `toLowerCase` field that is not a function...) but it expresses the intent of what you want to do more easily.

PR #91, opened 2016-04-15
feat(serve): Watch and serve all needed files

pixelastic:
Running `npm run serve` will watch for changes in the source files and
rebuild all `css` and `js` files accordingly in the `dist` folder. It
will also expose the same files on `http://localhost:8080/` (by
default, if port `8080` is already taken it will move to an available
one).

This will let developers use a specific branch/version of docsearch in
their projects. Just open a new tab and run `npm run serve` in the
docsearch repo, and you'll be able to include the scripts from the
running server in your HTML.

As files are automatically regenerated on disk whenever a change is
detected, you can also `require` files by targeting the `dist` folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,10 @@ npm run dev
 - `npm run build:docs` will create the `./docs/documentation.md` file from the
   `README.md` file, to be displayed on the website.
 - `npm run build` will run all three previous commands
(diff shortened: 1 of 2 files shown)


vvo:
This coveralls bot is triggering some good noise when nothing change :D

PR #97, opened 2016-05-22
Add keyboard bindings. Fixes #69

### LINKED ISSUE #69: Cmd+F / Ctrl+F keybinding
Would be great to jump directly to DocSearch using the default search key combinations.

maxiloc:


kaelig:
Awesome! I was actually going to do it manually so I'm glad that it's part of the script by default!

src/lib/DocSearch.js
```diff
@@ -184,6 +190,17 @@ class DocSearch {
     input.setVal('');
     window.location.href = suggestion.url;
   }
+
+  handleKeyboardShortcuts(input) {
+    $(document).keydown(function(e) {
+      if (false === input.is(':focus') && (e.which == 191 || e.which == 83)) {
```
pixelastic: I *think* you might fallback to `e.keyCode`  when `e.which` is not available. I kind of remember IE not following the `which` spec here. cc @vvo 

We should at least add a comment to explain what those magic numbers refers to. Or even let the user configure them in the options.

redox: > I think you might fallback to e.keyCode when e.which is not available. I kind of remember IE not following the which spec here. cc @vvo

`which` is OK since IE 9 (http://caniuse.com/#search=which)

vvo:
Maybe keyboard shortcuts should belong to autocomplete.js?

pixelastic:
Yep, we should add a configurable option to autocomplete.js to define the list of keys that will focus the search (maybe with a few defaults when we only say `keyboardShortcuts: true`).

maxiloc:
I guess I am going to redo it in autocomplete.js

PR #112, opened 2016-06-14
No result

maxiloc:


redox:
Can you please upgrade to `autocomplete.js@0.20.1`? It fixes a bug displaying the empty block when you clear the query.

PR #122, opened 2016-07-07
Update README for v2

ElPicador:


vvo:
I think you have other places with /1/

ElPicador:
Nice catch @vvo, I've fixed it

PR #156, opened 2016-11-09
style(camelCase): Fix camelCase linting issue

pixelastic:
Linting was failing because one variable was not camelCase


vvo:
tests are still failing for some reason (not lin)

vvo:
see #173

PR #164, opened 2017-01-24
feat(a11y): Hide separators to screen readers

kaelig:
This PR makes DocSearch more accessible to screen readers.

### DIFF
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -203,7 +203,7 @@ class DocSearch {
         utils.getHighlightedValue(hit, 'lvl4'),
         utils.getHighlightedValue(hit, 'lvl5'),
         utils.getHighlightedValue(hit, 'lvl6')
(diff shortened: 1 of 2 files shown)


kaelig:
More commits to come - tests in progress.

kaelig:
@vvo @maxiloc looks like this is all that is needed to make DocSearch accessible :)

kaelig:
@vvo @maxiloc any update on when this might get deployed?

maxiloc:
Yes going to work on it early next week when I come back to the office

maxiloc:
@kaelig It now released. I also made the same fix on docsearch v1 so every docsearch out there can get benefit from that 🎉

Thanks again for your work ❤️  ! This is a great addition to docsearch.

kaelig:
That's amazing, let's give it a few days to see how it goes and if everything is fine we can start announcing the update with the article :) thanks for your support!

PR #165, opened 2017-01-24
chore(build): Use include instead of exclude in webpack dev build

kaelig:
This makes local development easier with local builds of autocomplete.js.

### Before

`npm link autocomplete.js` throws an error because of a babel related thing

### After

`npm link autocomplete.js` works

### DIFF
--- a/webpack.dev.config.babel.js
+++ b/webpack.dev.config.babel.js
@@ -1,4 +1,4 @@
-import {join} from 'path';
+import path from 'path';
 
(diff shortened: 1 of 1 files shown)


vvo:
Thanks! What was the babel error?

kaelig:
Yep!

vvo:
> Yep!

I am sorry, I wanted to know what was the exact babel error (for reference, curiosity).

PR #167, opened 2017-02-07
docs(readme): update link to parameters

Haroenv:
The link used to point to the ruby explanations, but it didn’t exist at that id. Since it’s using js, I changed it to point to that point in the js documentation


vvo:
THX

PR #169, opened 2017-02-10
docs(readme): add installation instructions on npm

Haroenv:
docsearch is published on npm too, but it's a bit confusing that it's actually published as `docsearch.js`, so this should definitely be put somewhere in the README


maxiloc:
@Haroenv Thanks

PR #172, opened 2017-03-17
avoid the native Object's properties

### LINKED ISSUE #171: TypeError:  newCollection[key].push is not a function
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**

**If the current behavior is a bug, please provide all the steps to reproduce and a minimal
[JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install`
and `npm start`.**

**What is the expected behavior?**

https://eggjs.org/zh-cn/tutorials/index.html

untitled

monkindey:
**Summary**
fixed #171 

the given data type of docsearch hits might be conflict with the properties of the native Object, such as the constructor, so we need to do this check.

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -74,7 +74,9 @@ let utils = {
       if (typeof key === 'string') {
         key = key.toLowerCase();
       }
-      if (!newCollection[key]) {
+      // fix #171 the given data type of docsearch hits might be conflict with the properties of the native Object,
+      // such as the constructor, so we need to do this check.
+      if (!Object.prototype.hasOwnProperty.call(newCollection, key)) {
         newCollection[key] = [];
       }
       newCollection[key].push(item);


src/lib/utils.js
```diff
@@ -74,7 +74,9 @@ let utils = {
       if (typeof key === 'string') {
         key = key.toLowerCase();
       }
-      if (!newCollection[key]) {
+      // fix #171 the given data type of docsearch hits might be conflict with the properties of the native Object,
+      // such as the constructor, so we need to do this check.
+      if (!Object.prototype.hasOwnProperty.call(newCollection, key)) {
```
Haroenv: is this better than `key.hasOwnProperty(newCollection)`?

I don't think the comment is necessary though

vvo: see discussion here: https://github.com/algolia/docsearch/issues/171#issuecomment-286743620

Haroenv: sorry, makes sense

monkindey: Whether i should remove the comments inside the code?
And the travis failing is my wrong? If yes, I'll try my best to fix it.

Haroenv: the test failed because you based on `master`, not `develop`. You can change it in the GitHub UI just under your title if you click on edit

monkindey: Thanks a lot, I switch to `develop` branch.

monkindey:
Thanks to be merged.

PR #173, opened 2017-03-17
chore(stack update): update the whole docsearch stack

vvo:
In an effort to reduce our stack diversity I am moving docsearch to:
- latest eslint
- prettier
- webpack 2
- babel 6 (babel 7 is soon to be released, we were on babel 5)
- babel-preset-env (support: IE >=9, Two latest version of evergreen
browsers).
- jest


ElPicador:
Thanks @vvo. Here is a 🐼 as a gift.

PR #201, opened 2017-10-17
chore(docs): remove duplicate bootstrap gif

Haroenv:
this file isn't necessary, since it's not the real used file I think


vvo:
It's the one used in the README still

Haroenv:
Oops, didn't check there 🔥

PR #246, opened 2017-11-24
fix(autocompleteOptions): allow cssClasses options

davidohlin:
Let through options passed in `autocompleteOptions.cssClasses` and use `prefix: 'ds'` as default.

Right now, options being are overwritten [on these lines](https://github.com/algolia/docsearch/blob/master/src/lib/DocSearch.js#L76-L78).

Have a good weekend! ☀️


Haroenv:
Thanks for your contribution!

vvo:
All good

PR #262, opened 2018-02-05
fix(wording): fixed case of DocSearch

redox:


s-pace:
👍 
Welcome to the new naming standardisation aventure!

PR #283, opened 2018-04-24
chore(dev-env): fix dev env

vvo:
The developer environment was broken, could not run `npm run dev:docs`. I also upgraded safe dependencies and fixed the underlying security issues reported by GitHub.


vvo:
From now on you will need yarn to be installed (see https://yarnpkg.com/en/docs/install#alternatives-stable on how to install it)

docs/algolia-aerial/package.json
```diff
@@ -7,15 +7,15 @@
...
+    "extract-text-webpack-plugin": "^3.0.0",
+    "node-sass": "^4.5.3",
+    "sass-loader": "^6.0.6",
+    "webpack": "^3.10.0",
+    "webpack-auto-inject-version": "^1.0.0",
+    "webpack-livereload-plugin": "^0.11.0"
   },
   "name": "algolia-aerial",
-  "version": "1.2.6",
+  "version": "1.3.4",
```
Haroenv: How come this is inside this repo and not via dependency or cdn?

src/lib/__tests__/DocSearch-test.js
```diff
@@ -394,11 +389,14 @@ describe('DocSearch', () => {
 
   describe('handleSelected', () => {
     it('should change the location', () => {
+      const fakeWindow = { location: { href: null } };
```
Haroenv: Could also have written: `global.window = ...` and then the __window didn’t need to be added in the code

vvo: It still does not works because window or global.window is handled by jsdom no matter what so window.location.href fails completely

vvo: See https://github.com/jsdom/jsdom/issues/2112

Haroenv: any reason why you didn't do it like this: https://github.com/jsdom/jsdom/issues/2112#issuecomment-361195043?

vvo: It does require the actual implementation code to switch from href to assign, but we could try indeed if that's safe

pixelastic:
This is outdated with the new documentation website. Sorry for not having merged that sooner.

PR #339, opened 2018-05-24
Doc(content): Adding a FAQ

s-pace:
**Summary**

**Result**


s-pace:
@maxiloc @codeharmonics if you could had a super quick look that would be awesome

s-pace:
No rush, I will merge it on monday (reviews aren't critical)

maxiloc:
Looks good on the content part. I'll let Peter double check the language.

docs/source/documentation/4-docsearch-FAQ/1-customize-configuration-file.html.md.erb
```diff
@@ -0,0 +1,71 @@
...
+
+## What content do we recommend indexing?
+
+### Code blocks
+
+We **do not recommend** indexing **code tags** since code blocks are most likely
+very similar and redundant from one page to another.
+You can exclude code blocks by adding their matching selectors to `selectors_exclude`.
+
+A good practice would be to emphasize the meaningful underlying part of it
```
codeharmonics: I don't understand what is the "meaningful underlying part" of a code block? Maybe we can provide an example of a code block that has meaningful content?

docs/source/documentation/4-docsearch-FAQ/1-customize-configuration-file.html.md.erb
```diff
@@ -0,0 +1,71 @@
...
+    ]
+}
+```
+2. `level`: this value depends on the `level` of the record. A record's level is
+its **deepest level attribute not null**. `text` records have a weight of 0.
+3. `position`: This value is the position of the matching element within every
+picked up element along the original HTML flow. The **sooner** the record appears,
+the **higher** it will be ranked.
+
+You can override the way these elements are impacting the search thanks to `custom_settings`.
```
codeharmonics: I think you are saying that if you don't like, for example, our default `position` weight, then you can override it. But when I look at the example below, I don't see how that is being done. Can you add to the code snippet examples of values, like: 
```
weight.page_rank = ???
weight.level = ???
weight.position = ???

"custom_settings": {
+  "customRanking": [
+  "asc(weight.position)",
+    "desc(weight.page_rank)",
+    "desc(weight.level)"
+  ]
+}
```

docs/source/documentation/4-docsearch-FAQ/1-customize-configuration-file.html.md.erb
s-pace: The included part that you want to highlight:
 Example: the params of a code tag
<code>
blabla(<toIndex>something</toIndex>)
...
</code>

docs/source/documentation/4-docsearch-FAQ/1-customize-configuration-file.html.md.erb
s-pace: It will need some dev from our side but it was a way to show them that you could change it.
I may need to remove it.

s-pace:
Thx

PR #353, opened 2018-06-07
Fix incorrect text case for fastlane

revolter:
**Summary**

_fastlane_ is always written in lowercase 😃

**Result**


Haroenv:
whoops, thanks!

PR #493, opened 2018-10-09
Add precommit and prepush hooks that don't conflict

pixelastic:
I improved the way our git hooks were working to avoid conflict between the root repo and the ./docs folder.

Husky has been installed at the root and is configured through a `.huskyrc.js` file. It has a precommit and prepush hook defined. They will both check if there are changes made in the root or in the docs, and call the appropriate hook based on this. 

For example, when committing a change to `./docs/src/whatever.md`, it will run the `./docs/scripts/precommit` hook. When pushing changes that only affect the root directory, it will run `./scripts/prepush`, but when also pushing changes in `./docs`, it will also run `./docs/scripts/prepush`.

The current hooks are:
- Before pushing changes in the root, we run the tests and lint
- Before pushing changes in docs, we run the documentation linting
- Before committing in the docs, we try to auto-fix the linting issues of the modified files

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -1,0 +1,6 @@
+module.exports = {
+  hooks: {
+    'pre-commit': './scripts/precommit-dispatcher',
+    'pre-push': './scripts/prepush-dispatcher',
+  },
+};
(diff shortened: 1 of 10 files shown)


pixelastic:
Hey thanks, I didn't know about `lint-staged`. That would actually really simplify the bash dance of passing the list of files around!

s-pace:
Closing since this PR will not be updated for the foreseeable future

PR #500, opened 2018-10-11
[Issue template] Provide better guidance and redirection when opening an issue

s-pace:


.github/ISSUE_TEMPLATE.md
```diff
@@ -5,12 +5,29 @@
     - do at least one GitHub search in current issues, maybe your question is already here
 -->
 
-**Do you want to request a *feature* or report a *bug*?**
+**Is your issue related to your own DocSearch implementation?**
```
pixelastic: Good idea.

I would make sure the "if your issue is about your config, please post it on the config repo" is one of the first thing they see, though. Along with a link to directly open an issue there.

I'm not sure they need to provide the dropdown version they are using as they might not even know it. Maybe just ask for the `index_name` and live website url?

Not sure we need to add a link to a jsfiddle, that might not be very relevant here. Asking for screenshots instead would help

s-pace: Could we ask for material and offer both possibilities ?

Haroenv: You can have multiple templates to resolve this: https://github.com/algolia/vue-instantsearch/tree/master/.github/ISSUE_TEMPLATE

s-pace: Great. Thank you for chiming in @Haroenv . You are the best

pixelastic: Oh that's great. This would be a great way to dispatch users to the right template based on if their issue is for the relevance, for the docker, for the UI. :+1: to that

s-pace: ✅ https://github.com/algolia/docsearch/pull/500/commits/a6dca2d2824333c31624464133fc36207479ed27

PR #501, opened 2018-10-11
docs(css-helper): Add a page about a CSS helper to create a config

pixelastic:
I added a small CSS helper (to be loaded into any webpage with a user extension like Stylus) to quickly see the headers/text nodes. I've used it to help me create config faster instead of having to dig in the console.

Example


s-pace:
Chrome console is great IMHO too

Thanks

Can we use Xpath?

pixelastic:
> Chrome console is great IMHO too
Yep, it won't replace it, but allows to quickly see if a page uses a h1 or a h2 quickly (at least, it helped me)

> Can we use Xpath?
Unfortunately nope

s-pace:
Too bad. I will iterate on this point in order to explain some XPATH and CSS tricks I am doing a lot. It would be fun to explain that

PR #512, opened 2018-10-25
Doc/recommended configuration text update

s-pace:
Removing strident sentence


docs/src/recomended-configuration.md
```diff
@@ -1,31 +1,32 @@
 ---
 layout: two-columns
-title: Recommended recommendation
+title: Recommended configuration
```
pixelastic: Still need to rename the page to fix the typo as well :)

docs/src/recomended-configuration.md
```diff
@@ -1,31 +1,32 @@
...
 ---
 
 This is great news to know that you want to integrate DocSearch in your website.
 A good search experience is key to help your users discover your content.
 
-This section, [empowered by the details regarding how we build a DocSearch
-index][1], this section will give what is the requirements in order to have a
-great experience.
+This section, [empowered by the details about how do we build a DocSearch
```
pixelastic: Not sure `do` is needed in english here :grey_question:

docs/src/recomended-configuration.md
```diff
@@ -1,31 +1,32 @@
...
 
-- My website should have [an updated sitemap][2]. This is key in order to let us
-  know what should be updated. Do not worry, we will still crawl your website
-  and discover embedded hyperlinks to find your great content.
+- Your website should have [an updated sitemap][2]. This is key to let us know
+  what should be updated. Do not worry, we will still crawl your website and
+  discover embedded hyperlinks to find your great content.
 
-- Every pages needs to have her full context available. Using [metadata is
+- Every pages needs to have their full context available. Using [metadata is
```
pixelastic: I don't think it's very clear what you mean by `metadata is meaningful`? Are you talking about specific `<meta>` tags in the HTML? If so, we should list the one we'll use

docs/src/recomended-configuration.md
```diff
@@ -1,31 +1,32 @@
...
   meaningful][3].
 
-- Every `lvlx` DOM elements (matching your selectors) must have a unique `id` or
-  `name`. This will help the redirection to directly scroll down to the exact
-  place of the matching elements.
+- Every `lvlx` DOM elements (matching the selectors defined in your
+  configuration) must have a unique `id` or `name`. This will help the
+  redirection to directly scroll down to the exact place of the matching
+  elements.\[]
```
pixelastic: It seems there is an unused `\[]` at the end of the paragraph.

Also, doesn't it repeat what we already have in the `tips.md` page? I think we should really avoid having two pages with duplicate content and consolidate them into one page.

docs/src/recomended-configuration.md
```diff
@@ -1,31 +1,32 @@
...
 
 - Your website should not require some JavaScript rendering to generate the
   payload of your website (that-is-to-say your documentation). You can change
-  [the `user_agent` parameter][4] in order to do so.
+  [the `user_agent` parameter][4] to do so.
```
pixelastic: Are you sure it's not the `js_render` parameter?

s-pace: My idea was to promote a rendering based on the user_agent. I will remove it since I wasn't sure about this part

s-pace:
> I still think we have too much overlap between this page and the tips one.

Better be crystal clear and explain everything in the same place. Should be solve.

Please approve @pixelastic and I will merge it asap.

Thanks 🙏

s-pace:
replaced by #528

PR #515, opened 2018-11-01
Feat/jsbeautify

s-pace:
Adding js-beautify to format scss file

It will only impact the `./src`folder

New command:
`npm run beautify:scss`


s-pace:
Let's avoid this PR since it breaks variables with - in their name

Haroenv:
prettier works for css as well, you can use that instead

pixelastic:
@Haroenv does it work with SCSS as well?

Haroenv:
yes

pixelastic:
@s-pace that might me something to look at: having prettier run on both our CSS and JS codebase.

Haroenv:
https://github.com/yarnpkg/website/blob/master/package.json#L27, here's the example

pixelastic:
@Haroenv The command you linked is `yarn prettier --write '**/*.{js,md,scss}'`, how is that different from `prettier --write '**/*.{js,md,scss}'` (minus the `yarn` in front)? It will be run as an `npm`/`yarn` script anyway so `prettier` will be in the `$PATH` already, right?

PR #516, opened 2018-11-01
Feat/prettier

s-pace:
This PR introduces the use of prettier for scss & md files

solves #515


package.json
```diff
@@ -11,6 +11,8 @@
     "docs:serve": "cd ./docs && yarn serve",
     "docs:deploy": "cd ./docs && yarn deploy",
     "lint": "./scripts/lint",
+    "format:scss": "find ./src -type f -name '*.scss' | xargs yarn prettier --write ",
```
pixelastic:
```suggestion
    "format:scss": "prettier --write ./src/**/*.scss",
```
The `./**/*.scss` should grab all `.scss` files in directories and subdirectories.

package.json
```diff
@@ -11,6 +11,8 @@
...
+    "format:md": "yarn prettier --write '**/*.md'",
```
pixelastic:
```suggestion
    "format:md": "prettier --write '**/*.md'",
```
I don't think we need the `yarn` here.

s-pace:
Ready to merge @pixelastic ?

package.json
```diff
@@ -11,6 +11,8 @@
     "docs:serve": "cd ./docs && yarn serve",
     "docs:deploy": "cd ./docs && yarn deploy",
     "lint": "./scripts/lint",
+    "format:scss": "prettier --write ./src/**/*.scss",
+    "format:md": "prettier --write '**/*.md'",
```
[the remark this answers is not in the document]
s-pace: The use of prettier and eslint is not nice with this version (changes are the same with the VS extension but not with the module) so let's continue this way.

PR #526, opened 2018-11-06
Adding new logo

pixelastic:
I reverted to the previous way of displaying the logo, as a background
SVG image. Adding a new `<svg>` inline element to the markup could
break people implementation relying on our markup, so better play safe
and keep the markup the same.

I used the SVG available on our
https://www.algolia.com/press#resources page, but trimmed it through
SVGO (50% compression) before adding it to the background.

Before:
image
After:
image


s-pace:
I do not see any enhancement. The logo is not displayed well.

Could you use this svg instead please?
https://instantsearch-css.netlify.com/widgets/powered-by/

pixelastic:
> I do not see any enhancement. The logo is not displayed well.

What do you mean by "not displayed well"? It is using the same display as the previous one (same size and same alignment), but using the new coloring (no more gradient).

> Could you use this svg instead please?

Why do you think this version is better? I used the one on our resource page because it looks more like the previous logo we had (we keep the "Search by" test in black), so it's a softer change. And according to the design team, both versions are valid.

s-pace:
> What do you mean by "not displayed well"? It is using the same display as the previous one (same size and same alignment), but using the new coloring (no more gradient).

It looks blurry compared to the previous one

> Why do you think this version is better?

It will unify our `powered by logo`. We will reuse a work previously designed and tailored. Why should we not use it?

pixelastic:
> It looks blurry compared to the previous one

Ah yes, good call. The logo actually slightly changed dimensions. I updated:
image

> Why should we not use it?

Because we're upgrading live implementations of hundred of users, silently, to upgrade our branding. It has to be subtle, we don't want people to think we're using their website as our place to show advertisement for our logo. So we're using the logo version that is the closest to the one we had previously, to make the change smooth, but still following the new branding.

s-pace:
> Because we're upgrading live implementations of hundred of users, silently, to upgrade our branding. It has to be subtle, we don't want people to think we're using their website as our place to show advertisement for our logo. So we're using the logo version that is the closest to the one we had previously, to make the change smooth, but still following the new branding.

We are not doing that so fine.

pixelastic:
Checking with the IS team to see why they used an all-purple one

src/lib/templates.js
```diff
@@ -69,15 +69,9 @@ const templates = {
...
-            <path fill="#5468FF" d="M59.399.044h13.299a2.372 2.372 0 0 1 2.377 2.364v13.234a2.372 2.372 0 0 1-2.377 2.364H59.399a2.372 2.372 0 0 1-2.377-2.364V2.403A2.368 2.368 0 0 1 59.399.044z" />
-            <path fill="#FFF" d="M66.257 4.582c-2.815 0-5.1 2.272-5.1 5.078 0 2.806 2.284 5.072 5.1 5.072 2.815 0 5.1-2.272 5.1-5.078 0-2.806-2.279-5.072-5.1-5.072zm0 8.652c-1.983 0-3.593-1.602-3.593-3.574 0-1.972 1.61-3.574 3.593-3.574 1.983 0 3.593 1.602 3.593 3.574a3.582 3.582 0 0 1-3.593 3.574zm0-6.418V9.48c0 .076.082.131.153.093l2.377-1.226c.055-.027.071-.093.044-.147a2.96 2.96 0 0 0-2.465-1.487c-.055 0-.11.044-.11.104h.001zm-3.33-1.956l-.312-.31a.783.783 0 0 0-1.106 0l-.372.37a.773.773 0 0 0 0 1.1l.307.305c.049.05.121.038.164-.01.181-.246.378-.48.597-.698.225-.223.455-.42.707-.599.055-.033.06-.109.016-.158h-.001zm5.001-.806v-.616a.781.781 0 0 0-.783-.779h-1.824a.78.78 0 0 0-.783.78v.631c0 .071.066.12.137.104a5.736 5.736 0 0 1 1.588-.223c.52 0 1.035.071 1.534.207a.106.106 0 0 0 .131-.104z" />
-        </svg>
-    </a>
-  </div>
+    <div class="${footerPrefix}">
+      Search by <a class="${footerPrefix}--logo" href="https://www.algolia.com/docsearch">Algolia</a>
```
Haroenv: will this not have "search by" two times now?

pixelastic: Good catch but no, we hide the text with `font-size: 0` :)

pixelastic: The diff is actually compared to a previous PR that was merge too quickly. The "new" code is actually how it is currently in production (just a different background image)

s-pace: We have the text written within the DOM, might be an explanation

PR #542, opened 2018-11-20
feat(ctrl-click): Enable Ctrl-Click on suggestions

pixelastic:
This updates the underlying autocomplete.js to a version that passes
the selection context (click or keyboard) to the
`autocomplete:selected` event.

With the default `handleSelected`, and in the case of a click, it will
do nothing (as each suggestion already being a `<a href>`, it will
already follow the link), but manually change the location for
keyboard navigation.

When a custom `handleSelected` is set though, default browser clicks
on suggestions are disabled and only `handleSelected` is called. This
is to keep backward compatibility with all the live websites that
might already use `handleSelected`.

### DIFF
--- a/docs/src/behavior.md
+++ b/docs/src/behavior.md
@@ -24,24 +24,34 @@ docsearch({
 
 ## `handleSelected`
 
-This method is called when a suggestion is selected. By default, DocSearch will
-redirect the browser to the results page at the related anchor, but you can
-override it to add your own behavior.
+This method is called when a suggestion is selected (either with a click or
+keyboard). By default, DocSearch will redirect the browser to the results page
+at the related anchor, but you can override it to add your own behavior.
 
-The method is called with three arguments:
+The method is called with the following arguments:
 
 - `input`, a reference to the search `input` element. It comes with the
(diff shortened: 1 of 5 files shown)


docs/src/behavior.md
```diff
@@ -24,24 +24,34 @@ docsearch({
...
 
-- `event`, the actual event triggering the selection. This can come from a click
-  or a keyboard navigation.
+- `event`, the actual event triggering the selection.
 
-- `suggestion`, the object representing the current selection.
+- `suggestion`, the object representing the current selection. Contains a `.url`
+  key representing the destination.
+
+- `datasetNumber`, this should always equal `1` as DocSearch is searching
```
s-pace: IMHO we should not document this input if it is not useful with our UI which is already tight with one index only.

s-pace:
@pixelastic can you release it with a new version please?

docs/src/behavior.md
pixelastic: The issue is that it's the 3rd parameter out of 4, that was already there before. If I don't document it, it make it look like there are only 3 parameters. And we need to keep it to not break backward compatibility, so I unfortunately have to keep it that way :/

PR #543, opened 2018-11-20
Omit redundant `type="text/javascript"`

mathiasbynens:
**Summary**

This removes some redundant code from the recommended HTML snippet. See <https://mathiasbynens.be/notes/html5-levels#type-attributes>.

**Result**

This is valid HTML. `text/javascript` is the default `type` for `<script>` elements per spec, and this is implemented by all browsers.

### DIFF
--- a/scripts/playground.html
+++ b/scripts/playground.html
@@ -20,8 +20,8 @@
       </div>
     </div>
     <!-- at the end of the BODY -->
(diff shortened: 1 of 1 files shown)


s-pace:
Thanks for the contribution

s-pace:
As a really small thanks @mathiasbynens ,

Would you be interest in having docsearch for https://mathiasbynens.be/notes/ ?

We would be to help this website. It is really great.

Let us know

PR #544, opened 2018-11-21
ci: use vm-based travis

Haroenv:
Let's see if this breaks anything, since it will become required soon: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


Haroenv:
looks like it's working, yay ☯️

PR #550, opened 2018-11-27
fix(security): Fix dependencies to avoid infected event-stream

pixelastic:
More info: https://github.com/dominictarr/event-stream/issues/116


s-pace:
COuld you check netlify too please?

pixelastic:
Netlify fixed. It was missing a `yarn install`, that's why.

PR #553, opened 2018-11-27
docs(hub): Plug apply form to the hub

pixelastic:
Plugs the apply form to the new hub.

Live demo available on: https://5c000c78c6aed67c65fa0ac6--docsearch.netlify.com/

Adds the repository url field:
image

It displays error messages as returned by the API:
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,7 +4,8 @@ module.exports = {
   globals: {
     docsearch: true,
     Glide: true,
-    Popper: true
+    Popper: true,
+    grecaptcha: true
   },
   rules: {
     'no-console': 0,
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ yarn-error.log
 dist-es5-module/
 tmp
 docs/.textlintcache
+.netlify
--- a/docs/config.json
+++ b/docs/config.json
@@ -7,7 +7,8 @@
     "repo": "https://github.com/algolia/docsearch"
   },
   "placeholders": {
-    "docSearchJSVersion": "2"
+    "docSearchJSVersion": "2",
+    "recaptchaKey": "6Le61n0UAAAAAGruCSV0p8Hl7zyQo9F3fzaSmxwJ"
   },
   "sidebar": [
     {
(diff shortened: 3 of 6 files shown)


pixelastic:
@Shipow Let me know when it's ok to merge and have it live

pixelastic:
I added an invisible recaptcha. When the form will be submitted it will check if the user is a human or a bot. Most of the time it will be invisible, and we'll submit the form as usual (there is still a backend check done in https://github.com/algolia/docsearch-hub-backend/pull/3). If it has reasonable doubts about the user not being human, it will prompt a captcha asking to click on traffic light / cars / etc.

You can see it live on https://5c000c78c6aed67c65fa0ac6--docsearch.netlify.com/ but it won't work until https://github.com/algolia/docsearch-hub-backend/pull/3 is merged.

I updated the TOS because we need to say that we're using recaptcha:
image

Also, in the case the backend validation fails, we display the following message:
image
^ this should happen only when the token is invalid (ie. someone is trying to brute-force the form), but just to be sure we're not missing on any genuine request, it's better to provide this fallback.

pixelastic:
Issues raised by @s-pace:

- githubUrl is marked as optional, but is actually required. @Shipow I suggest we ship this version with it being optional, and we'll make it mandatory when we'll have gitlab support in the backend. WDYT?
- Having an error with a hard refresh with pre-filled fields: no message displayed

Shipow:
I'd rather put the github repo mandatory and put a contact link for the rare projects that don't have one. I'll add Gitlab <3 support quickly then.

pixelastic:
Updated to work with https://github.com/algolia/docsearch-hub-backend/pull/14

- Make `repoUrl` mandatory instead of `githubUrl` optional
- Expect errors returned as array of objects with `.field`, `.message` and `.code`.
- Also catches potential errors form the backend and display the "contact us by email" as a fallback

pixelastic:
@Shipow I'm still having a CORS issue with the deploy preview. I thought we managed to fixed it at some point, but now it's back.

I tested locally, by loading the form on `localhost` and the API on `localhost.net` and I can correctly do a call from one to the other. But when on Netlify, it still fails with `No 'Access-Control-Allow-Origin' header is present on the requested resource.`. I have absolutely no idea how to fix that, considering that the initial `OPTIONS` requests correctly returns a `Access-Control-Allow-Origin: *`

s-pace:
Closing since the form will not be updated for the foreseeable future

Issue #555, opened 2018-11-28
QueryID needed from query data for Algolia click analytics


harrison-symes:
**Is your issue related to your own DocSearch implementation?**

For custom implementation with Algolia Click Analytics (https://www.algolia.com/doc/guides/insights-and-analytics/analytics-overview/#analytics---implementation-overview) the QueryID is needed from the query request data response. 

Within `src/lib/DocSearch.js`, the `getAutocompleteSource` function performs a search query from the Algolia client. 

``` js 
 this.client
        .search([
          {
            indexName: this.indexName,
            query,
            params: this.algoliaOptions,
          },
        ])
        .then(data => {
          // This data variable here will contain the needed information
          let hits = data.results[0].hits;
          if (transformData) {
            hits = transformData(hits) || hits;
          }
          callback(DocSearch.formatHits(hits));
        });
```

The QueryId can be found within the response data, but the Docsearch API does not currently allow the developer to access this full response object, only the `data.results[0].hits` through various means.

I propose that an optional `queryDataCallback` argument for the Docsearch instance be implemented that will call back this full response `data` object so a developer may observe and interact with all response information, rather than just `hits`

harrison-symes:
https://github.com/algolia/docsearch/pull/556

Haroenv:
closed by #556

PR #556, opened 2018-11-28
implement query data callback method into src

harrison-symes:
**Summary**
https://github.com/algolia/docsearch/issues/555

**Result**

``` js
// Usage of new method
docsearch({
    ...config,
    inputSelector: "#docsearch-input",
    dataCallback: data => formatData(data)
    ...
});
---

const clickObject = {
    objectID,
    position,
    queryID //queryID of inner query, can now retrieve from within DocSearch instance
};
        
// Algolia Analytics library
aa("click", clickObject); //send the click data to algolia
```

### DIFF
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -36,6 +36,7 @@ class DocSearch {
     appId = 'BH4D9OD16A',
     debug = false,
     algoliaOptions = {},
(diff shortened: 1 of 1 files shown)


Haroenv:
Thanks for the fast turnaround! ⚡️

s-pace:
Thanks a lot

PR #563, opened 2018-12-13
docs(deploy-preview): Re-enable deploy previews for the doc

pixelastic:
This re-enable deploy-previews through Netlify. It will check if
changes where made to the ./docs folder and build + deploy the preview
if so. Otherwise, it will deploy a dummy preview.

This also removes the netlify script to deploy in production whenever
we push to master, as this is now handled by Travis.


scripts/netlify/changes-in-docs-since-master
```diff
@@ -0,0 +1,15 @@
...
+# Compare current branch with master to see if there are changes in the docs/
+# folder
+set -x
+
+# Remote urls are not always set on Netlify
+git remote set-url origin https://github.com/algolia/docsearch.git
+git fetch
+
+CHANGES_IN_DOCS="$(git diff --name-only origin/master | grep '^docs/')"
+if [ "$CHANGES_IN_DOCS" = "" ]; then
```
s-pace:
```suggestion
if [ $CHANGES_IN_DOCS = "" ]; then
```

s-pace: Not tested but I think it should be set this way

scripts/netlify/is-deploy-preview
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+if [ "$CONTEXT" = "deploy-preview" ]; then
```
s-pace:
```suggestion
if [ $CONTEXT = "deploy-preview" ]; then
```

scripts/netlify/changes-in-docs-since-master
pixelastic: If you don't set the quotes, it will fail if for some reason `$CHANGES_IN_DOCS` contains spaces or glob characters (like `*`). In that case it _shouldn't_ happen, but it's best practice in bash to always quote strings you want to test.

I actually use [shellcheck](https://www.shellcheck.net/) in my IDE to warn me about those issues directly

s-pace: Thank you for the explanation. I have noticed it and removed this review.
Thanks for the TIP

PR #569, opened 2018-12-19
update documentation as algolia/docsearch-scraper#431

s-pace:
**Summary**

**Result**

### DIFF
--- a/docs/src/config-file.md
+++ b/docs/src/config-file.md
@@ -606,11 +606,19 @@ DocSearch to index all your content.
 
 You can override the user agent used to crawl your website. By default, this
 value is:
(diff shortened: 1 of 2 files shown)


docs/src/config-file.md
```diff
@@ -606,11 +606,19 @@ DocSearch to index all your content.
 
 You can override the user agent used to crawl your website. By default, this
 value is:
-`'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_1) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/70.0.3538.110 Safari/537.36'`
+
+    Algolia DocSearch Crawler
```
pixelastic:
```suggestion
     `Algolia DocSearch Crawler`
```

pixelastic: I think the current formatting will put the UA in its own code block. Wrapping in backticks should correctly put it inline.

s-pace: I let you merge if you agree on that

PR #571, opened 2018-12-20
Update/doc

s-pace:
- Fix header
- Fix typos
- Introduce meta tags

### DIFF
--- a/docs/src/behavior.md
+++ b/docs/src/behavior.md
@@ -3,17 +3,19 @@ layout: two-columns
 title: Dropdown Behavior
 ---
 
-Our JS library `docsearch.js` is a wrapper of the [autocomplete.js][1] library.
-This library will listen to every keystrokes in the search input, query Algolia,
-and display the results in a dropdown. Everything is already configured for you
-to work with DocSearch. Our UI library also exposes configuration options you
-can use to go even further. You will discover Algolia.
+Our JS library `docsearch.js` is a wrapper of the [Algolia autocomplete.js][1]
+library. This library will listen to every keystrokes type in the search input,
+query Algolia, and display the results in a dropdown. Everything is already
+configured for you to work with DocSearch. Our UI library also exposes
+configuration options you can use to go even further. You will discover Algolia
+out of the box for documentation. Let's start the **learn as you type**
+experience.
 
 ## `appId`
 
-If you're [running the DocSearch crawler on your own][2], you'll need to define
-your own application ID using the `appId` key. If you're using the free hosted
-version, you don't need to add anything.
+Only required if you're [running the DocSearch crawler on your own][2]. It
+defines your own application ID using the `appId` key. If you're using the free
+hosted version, you don't need to consider this parameter.
 
 ```javascript
 docsearch({
@@ -26,8 +28,9 @@ docsearch({
 
 This method is called when a suggestion is selected (either from a click or a
 keystroke). By default, DocSearch will display links redirecting to the results
-page, at the related anchor. You can override it to add your own behavior.
-Please note that you can already open a new tab thanks to the `CMD/CTRL + Click`
+page, at the related position (using anchor). You can override results (hit) to
+add your own behavior. Please note that you can already open a new tab thanks to
+the `CMD/CTRL + Click` action.
 
 The method is called with the following arguments:
 
@@ -36,11 +39,11 @@ The method is called with the following arguments:
 
 - `event`, the actual event triggering the selection.
 
-- `suggestion`, the object representing the current selection. Contains a `.url`
-  key representing the destination.
+- `suggestion`, the object representing the current selection. It contains a
+  `.url` key representing the destination.
 
-- `datasetNumber`, this should always equal `1` as DocSearch is searching into
-  one dataset at a time. You can ignore this attribute.
+- `datasetNumber`, this should always be equal `1` as DocSearch is searching
+  into one dataset at a time. You can ignore this attribute.
 
 - `context`, additional information about the selection. Contains a
   `.selectionMethod` key that can be either `click`, `enterKey`, `tabKey` or
@@ -50,7 +53,13 @@ The method is called with the following arguments:
 docsearch({
   […],
   handleSelected: function(input, event, suggestion, datasetNumber, context) {
-    // Default implementation is as follow:
+    // Do nothing if click on the suggestion, as it's already a <a href>, the
+    // browser will take care of it. This allow Ctrl-Clicking on results and not
+    // having the main window being redirected as well
+    if (context.selectionMethod === 'click') {
+      return;
+    }
+
     input.setVal('');
     window.location.assign(suggestion.url);
   }
@@ -59,9 +68,9 @@ docsearch({
 
 ## `queryHook`
 
-This method will be called on every keystroke to transform the typed keywords
-before querying Algolia. By default, it does not do anything, but we provide
-this hook for you to add your own logic if needed.
+This method is called on every keystroke to transform the typed keywords before
+querying Algolia. By default, it does not do anything, but we provide this hook
+for you to add your own logic if needed.
 
 ```javascript
 docsearch({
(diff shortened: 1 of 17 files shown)


docs/src/behavior.md
```diff
@@ -3,17 +3,19 @@ layout: two-columns
 title: Dropdown Behavior
 ---
 
-Our JS library `docsearch.js` is a wrapper of the [autocomplete.js][1] library.
-This library will listen to every keystrokes in the search input, query Algolia,
-and display the results in a dropdown. Everything is already configured for you
-to work with DocSearch. Our UI library also exposes configuration options you
-can use to go even further. You will discover Algolia.
+Our JS library `docsearch.js` is a wrapper of the [Algolia autocomplete.js][1]
+library. This library will listen to every keystrokes type in the search input,
```
pixelastic:
```suggestion
 library. This library will listen to every keystrokes typed in the search input,
```

docs/src/behavior.md
```diff
@@ -3,17 +3,19 @@ layout: two-columns
...
+query Algolia, and display the results in a dropdown. Everything is already
+configured for you to work with DocSearch. Our UI library also exposes
+configuration options you can use to go even further. You will discover Algolia
+out of the box for documentation. Let's start the **learn as you type**
+experience.
```
pixelastic: `You will discover Algolia out of the box for documentation. Let's start the **learn as you type** experience.`

I don't really understand these sentences, they sound weird. "You will discover" and "Let's start the XXX experience" sound like something coming out of a TV advert IMO, I'm not a fan

docs/src/behavior.md
```diff
@@ -3,17 +3,19 @@ layout: two-columns
...
 
 ## `appId`
 
-If you're [running the DocSearch crawler on your own][2], you'll need to define
-your own application ID using the `appId` key. If you're using the free hosted
-version, you don't need to add anything.
+Only required if you're [running the DocSearch crawler on your own][2]. It
```
pixelastic:
```suggestion
 Required if you're [running the DocSearch crawler on your own][2]. It
```

pixelastic: Also, the next sentence seems to be repeating itself

docs/src/behavior.md
```diff
@@ -26,8 +28,9 @@ docsearch({
 
 This method is called when a suggestion is selected (either from a click or a
 keystroke). By default, DocSearch will display links redirecting to the results
-page, at the related anchor. You can override it to add your own behavior.
-Please note that you can already open a new tab thanks to the `CMD/CTRL + Click`
+page, at the related position (using anchor). You can override results (hit) to
```
pixelastic: I don't get the `you can override results (hit)` part. First I think we need to be consistent and either use records or hit, but not both. Second, it's not the actual records you can overwrite to change the behavior, but what is happening when you click on them (ie, it's a function you change, not a data object).

docs/src/behavior.md
```diff
@@ -36,11 +39,11 @@ The method is called with the following arguments:
...
 - `event`, the actual event triggering the selection.
 
-- `suggestion`, the object representing the current selection. Contains a `.url`
-  key representing the destination.
+- `suggestion`, the object representing the current selection. It contains a
+  `.url` key representing the destination.
 
-- `datasetNumber`, this should always equal `1` as DocSearch is searching into
-  one dataset at a time. You can ignore this attribute.
+- `datasetNumber`, this should always be equal `1` as DocSearch is searching
```
pixelastic: Either `will always equal something`, or `will always be equal to something`. Here you either are missing a `to`, or have a `be` that shouldn't be here

docs/src/dropdown.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: two-columns
-title: Dropdown
+title: Dropdown Search-UI
```
pixelastic: Why the dash?

docs/src/how-do-we-build-an-index.md
```diff
@@ -5,54 +5,57 @@ title: How do we build a DocSearch index?
...
 
-### Everything starts from your page
+## Everything starts from your page
 
 ![1st step][2] {mt-2}
 
-_Note: We would rather avoid useless js rendering. All useful information should
-be available without any client-side rendering. You can use `curl` in order to
-have a better idea of how it would look without any client-side rendering._
+_Note: We would rather avoid useless JavaScript rendering. All useful
```
pixelastic: I feel this comment is confusing. What is the intent behind it? What do you want to tell the reader? As a reader, I'm not sure what I should do with this information and I don't see how it makes me better understand how the crawler works.

docs/src/how-do-we-build-an-index.md
```diff
@@ -5,54 +5,57 @@ title: How do we build a DocSearch index?
...
 `min_indexed_level` is set to `0`)
 
-### We pile up the elements based on the current temporary record
+## We pile up the elements based on the current temporary record
 
 ![5th step][6] {mt-2}
 
 Base on the position within the flow, we nest elements as much as possible to
-keep the contextual relevancy.
+keep the context and incerease the relevancy.
```
pixelastic:
```suggestion
 keep the context and increase the relevancy.
```

docs/src/recommended-configuration.md
```diff
@@ -1,17 +1,17 @@
 ---
 layout: two-columns
-title: Recommended configuration
+title: Recommendation
```
pixelastic: What about `Best Practices`?

docs/src/recommended-configuration.md
```diff
@@ -62,6 +62,31 @@ The biggest blue element will be you `docSearch-content` container. Every
 selectors outside this element will be `global`. Every selectors appear in the
 same order than their `lvl` along the HTML flow.
 
+### Introduces global information as meta tags
+
+Our crawler automatically extract information from our DocSearch specific meta
+taf:
```
pixelastic:
```suggestion
 tags:
```

docs/src/recommended-configuration.md
```diff
@@ -62,6 +62,31 @@ The biggest blue element will be you `docSearch-content` container. Every
...
+
+```html
+<meta name="docsearch:language" content="en">
+<meta name="docsearch:version" content="1.0.0">
+```
+
+The `content` value of the meta will be added to every records extracted from
+the page. Giving that the name is `docsearch:$NAME`, `$NAME` will be set as a
```
pixelastic:
```suggestion
 the page. Given that the name is `docsearch:$NAME`, `$NAME` will be set as a
```

docs/src/recommended-configuration.md
```diff
@@ -62,6 +62,31 @@ The biggest blue element will be you `docSearch-content` container. Every
...
+attribute in every of them. Its value will be its related `content`value. You
```
pixelastic:
```suggestion
 attribute in each of them. Its value will be its related `content` value. You
```

docs/src/recommended-configuration.md
```diff
@@ -62,6 +62,31 @@ The biggest blue element will be you `docSearch-content` container. Every
...
+can then transform them as [`facetFilter`][6] to let you filter over it.
```
pixelastic:
```suggestion
 can then transform them as [`facetFilters`][6] to filter over them.
```

docs/src/recommended-configuration.md
```diff
@@ -62,6 +62,31 @@ The biggest blue element will be you `docSearch-content` container. Every
...
+Look at this example with our vuepress integration:
+
+```json
+"custom_settings": {
+    "attributesForFaceting": [
+      "lang"
+    ]
+  }
+```
+
```
pixelastic: You're mentioning `facetFilters` (for the front) above, then here using `attributesForFaceting` (for the settings). I think it needs a bit more clarification.

docs/src/who-can-apply.md
```diff
@@ -57,4 +57,5 @@ criteria:
 [1]: ./apply.html
 [2]: ./run-your-own.html
 [3]: https://www.algolia.com/doc/api-reference/
-[4]: ./integrations.html
+
+E[4]&#x3A; ./integrations.HTML
```
pixelastic: Seems like something is broken here

docs/src/behavior.md
s-pace: I think it is fine to have our catchphrase here and explain we are providing Algolia out of the box

docs/src/behavior.md
s-pace: That's right, I think using both term here is nice to introduce Algolia global documentation

PR #575, opened 2018-12-25
fix 404 page caused by .HTML

loveky:
**Summary**

`.HTML` leads to a 404 error in Chrome:

snip20181226_1

**Result**


s-pace:
Thanks

PR #590, opened 2019-01-22
Update handleSelect documentation

### LINKED ISSUE #588: Ctrl+Click on suggestion should open result in new tab, even if using custom `handleSelected`
**Do you want to request a _feature_ or report a _bug_?**
bug

**Steps to reproduce**
Assume the following code (copied from [documentation](https://community.algolia.com/docsearch/behavior.html#handleselected)):
<pre><code>docsearch({
  […],
  handleSelected: function(input, event, suggestion, datasetNumber, context) {
    // Do nothing if click on the suggestion, as it's already a &lt;a href&gt;, the
    // browser will take care of it. <b>This allow Ctrl-Clicking on results and not
    // having the main window being redirected as well</b>
    if (context.selectionMethod === 'click') {
      return;
    }

    ...
  }
});
</code></pre>
Then search for anything and Ctrl+Click on suggestion.

**What is the current behavior?**
Link is opened in current tab.

**What is the expected behavior?**
Link is opened in new tab.

**Proposed solution**
The problem is that if custom `handleSelected` is used, then [algolia prevents all clicks on `.algolia-autocomplete`](https://github.com/algolia/docsearch/blob/6e933604fb86ac22e6d71309fa29f600ab022fae/src/lib/DocSearch.js#L108-L113). I propose to prevent click only when it is normal click (not Ctrl+Click), or even ignore custom `handleSelected` when it is Ctrl+Click (or add option for one of those behaviours)

**If the current behavior is a bug, please provide all the steps to reproduce and a minimal
[JSFiddle][3] example or a repository on GitHub that we can `npm install`
and `npm start`.**
Sorry, I can't create minimal reproduction, because I can't find any instructions for creating it (maybe you will add such instructions, for example premade codepen template (with entered APPLICATION_ID), so that someone who wants to report a bug will just fork this codepen...)

[1]: https://github.com/algolia/docsearch-configs
[2]: https://community.algolia.com/docsearch/config-file.html
[3]: (https://jsfiddle.net/)

Thank you.

s-pace:
resolves #588 and avoid wrong comment

### DIFF
--- a/docs/src/behavior.md
+++ b/docs/src/behavior.md
@@ -64,7 +64,7 @@ docsearch({
 ```
 
 When a custom `handleSelected` is set, default browser clicks on suggestions are
-disabled and only `handleSelected` is called.
+turned off and only `handleSelected` is called.
 
 ## `queryHook`
 
--- a/docs/src/behavior.md
+++ b/docs/src/behavior.md
@@ -53,18 +53,18 @@ The method is called with the following arguments:
 docsearch({
   […],
   handleSelected: function(input, event, suggestion, datasetNumber, context) {
-    // Do nothing if click on the suggestion, as it's already a <a href>, the
-    // browser will take care of it. This allow Ctrl-Clicking on results and not
-    // having the main window being redirected as well
+
     if (context.selectionMethod === 'click') {
+      //Your own logic
       return;
     }
 
-    input.setVal('');
-    window.location.assign(suggestion.url);
   }
 });
 ```
+
+When a custom `handleSelected` is set, default browser clicks on suggestions are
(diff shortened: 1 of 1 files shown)


docs/src/behavior.md
```diff
@@ -53,19 +53,19 @@ The method is called with the following arguments:
...
       return;
     }
 
-    input.setVal('');
-    window.location.assign(suggestion.url);
   }
 });
 ```
 
+When a custom `handleSelected` is set, default browser clicks on suggestions are
```
francoischalifour: This phrase is not necessary IMO if you read the first paragraph of the method description.

PR #632, opened 2019-03-13
docs(readme): remove ToC

Haroenv:
don't merge, still need to remove the tool itself too I think (will do if approved)

**Summary**

**Result**


Haroenv:
bunch of removed transitive dependencies 🙌

PR #648, opened 2019-04-03
hashRouter anchor

nikomal:
**Summary**

if a website handle route by HashRouter.  such as: [react-router](https://github.com/ReactTraining/react-router).  then the url will has char '#', but it isn`t mean an anchor link. so in search result, the anchor will not append in url

**Result**
exclude the string like '#/*'.  if url has '#/', it will be recognized as a route, not an anchor.

### DIFF
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -305,7 +305,7 @@ class DocSearch {
   static formatURL(hit) {
     const { url, anchor } = hit;
     if (url) {
-      const containsAnchor = url.indexOf('#') !== -1;
+      const containsAnchor = !!url.match(/#[^\/]/);
       if (containsAnchor) return url;
       else if (anchor) return `${hit.url}#${hit.anchor}`;
       return url;


s-pace:
👋 @nikomal 

Could you provide us a live example of the issue this PR will solve? We would try to reproduce it in order to fully understand your issue.

It seems that the current implementation works with website handling route by HashRouter. You can have a look on the [signal documentation](http://signal.agworld.com/) for example.

nikomal:
👋 @s-pace

nikomal:
👋 @s-pace 

https://sheinsight.github.io/shineout/1.2.x/en#/  just like this link. 

the website use hashRouter and docsearch.js, but in the search result. because of the url has '#'. so the result link does not has the anchor.  this is special scene. it need both hashRouter and anchor link.

s-pace:
Thanks for the details. Given the way an URL is built and how a fragment is built from an URL, we consider that an URL generally contains one impactful `#`.

However, what you aim to do is definitely doable thanks to the [transformdata method](https://community.algolia.com/docsearch/behavior.html#transformdata). You can use it in this way:

```js
 docsearch({
    apiKey: '25626fae796133dc1e734c6bcaaeac3c',
    indexName: 'docsearch',
    inputSelector: '#q',
    transformData: function(hits) {
      // modify hits
      hits.map(hit => hit.url = hit.url + "#" + hit.anchor)
      return hits;
    },
    debug: false // Set debug to true if you want to inspect the dropdown
  });
```

s-pace:
Closing this PR since my previous comments solved the issues raised. 
@nikomal feel free to re-open it if needed.

Issue #418, opened 2018-08-20
Accessibility issue


oliviertassinari:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

The following lighthouse test is failing:

![capture d ecran 2018-08-21 a 00 26 43](https://user-images.githubusercontent.com/3165635/44370065-f085ec80-a4d8-11e8-9b9f-7e117ade6308.png)

You can reproduce it on [Bootstrap](http://getbootstrap.com/docs/4.1/getting-started/introduction/) or [Material-UI](https://material-ui.com/) with Chrome 68.

**What is the expected behavior?**

💯 % score

s-pace:
Thank you for sharing. Do you need a fix asap or can it wait for the v3?

oliviertassinari:
@s-pace as you see fit.

Lausselloic:
Hello,

Looks like aria-label is not saved on destroy. On bootstrap there's an "aria-label" attribute on the search input, but when checking on docserach.js v2.6.2

```
// store the original values of the attrs that get modified
  // so modifications can be reverted on destroy
  $input.data(attrsKey, {
    'aria-autocomplete': $input.attr('aria-autocomplete'),
    'aria-expanded': $input.attr('aria-expanded'),
    'aria-owns': $input.attr('aria-owns'),
    autocomplete: $input.attr('autocomplete'),
    dir: $input.attr('dir'),
    role: $input.attr('role'),
    spellcheck: $input.attr('spellcheck'),
    style: $input.attr('style'),
    type: $input.attr('type')
  });
```
quick-fix will be to add `aria-label` to the saved attributes?

endiliey:
It will be great to have this so that we can improve accessibility 😉

oliviertassinari:
I have patched the problem with: https://github.com/mui-org/material-ui/blob/e4ef5bf1adc6df6888a947919124302ed29d3343/docs/src/modules/components/AppSearch.js#L50-L67.

endiliey:
Thanks @oliviertassinari. I've sent a PR for it but I guess I'll use your solution for now.

cc @s-pace

endiliey:
We can close this out. I've resolved it after upgrading to 2.6.3

s-pace:
Thanks again @endiliey, version released 🚀

oliviertassinari:
Do we have to wait a week to see the CDN cache expire?

s-pace:
[...]

PR #679, opened 2019-05-10
fix: search input a11y issue

endiliey:
**Summary**

#418 

The search input does not have `aria-label`. I have found the underlying cause, it is because we are not passing `ariaLabel` options to autocomplete.js
See https://github.com/algolia/autocomplete.js/#global-options

> `ariaLabel` - An optional string that will populate the aria-label attribute.

My implementation is
- Check autoCompleteOptions for `ariaLabel`
- Otherwise, check original input selector `aria-label`
- Otherwise, use `search input` as aria-label

**Test Plan**

Here is my input

```html
<input
  id="search_input_react"
  type="search"
  placeholder="Search"
  aria-label="This is my aria-label from endiliey"
/>
```

**Before**
No aria-label
image

**After**
aria-label exist
aria-label working

### DIFF
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -80,6 +80,7 @@ class DocSearch {
       this.autocompleteOptions.cssClasses || {};
     this.autocompleteOptions.cssClasses.prefix =
       this.autocompleteOptions.cssClasses.prefix || 'ds';
+   this.autocompleteOptions.ariaLabel = this.autocompleteOptions.ariaLabel || $(this.input)[0].getAttribute('aria-label') || 'search input';
 
 
     this.isSimpleLayout = layout === 'simple';


s-pace:
👏 @endiliey , you are now a DocSearch collaborator 🚀

s-pace:
Reverting since:
```js
DocSearch › constructor › should allow customize algoliaOptions without loosing default options

    TypeError: Cannot read property 'getAttribute' of undefined

      81 |     this.autocompleteOptions.cssClasses.prefix =
      82 |       this.autocompleteOptions.cssClasses.prefix || 'ds';
    > 83 |    this.autocompleteOptions.ariaLabel = this.autocompleteOptions.ariaLabel || $(this.input)[0].getAttribute('aria-label') || 'search input';
         |                                                                               ^
      84 |
      85 |
      86 |     this.isSimpleLayout = layout === 'simple';

```

s-pace:
Do you have an idea @endiliey ?

endiliey:
$(this.input)[0] might be undefined. I should've used a safe getter like 
```js
$(this.input)[0] && $(this.input)[0].getAttribute('aria-label')
```

s-pace:
Ok, feel free to submit a new PR :)

endiliey:
@s-pace sorry for the delay. was doing something else

see https://github.com/algolia/docsearch/pull/687

Note that autocomplete.js must be >= v0.36 because `autoCompleteOptions.ariaLabel` is not available in < 0.36

PR #773, opened 2019-07-31
fix(templating): replace Hogan.js with ES6 literals

LoneRifle:
**Summary**
Hogan.js uses Function objects, which violates CSP policies which do not
add the `unsafe-eval` directive. Given the relatively minor use in the
codebase, replace with ES6 literals and remove the dependency

- Replace Hogan with direct interpolation of data into string template
- Rework/remove tests that make reference to Hogan.js, since not used
- Remove Hogan.js from dependencies

See algolia/instantsearch.js#2868 for further information

Inspect bcc440a for the actual changes - de-linting/prettier was run on affected files before changes committed

**Result**
Tested on local machine. Note the lack of references to Hogan

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
   "dependencies": {
     "algoliasearch": "^3.24.5",
     "autocomplete.js": "0.36.0",
-    "hogan.js": "^3.0.2",
     "request": "^2.87.0",
     "stack-utils": "^1.0.1",
     "to-factory": "^1.0.0",
--- a/src/lib/DocSearch.js
+++ b/src/lib/DocSearch.js
@@ -1,4 +1,3 @@
-import Hogan from 'hogan.js';
 import algoliasearch from 'algoliasearch/lite';
 import autocomplete from 'autocomplete.js';
 import templates from './templates';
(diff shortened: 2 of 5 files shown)


Haroenv:
This is very interesting, thanks! I'm not a maintainer of docsearch, but don't we expose the templates to the users? If so this would be a breaking change

Haroenv:
Bundlesize impact:

pre: 46.6 kB gz
post: 42.5 kB gz

LoneRifle:
> This is very interesting, thanks! I'm not a maintainer of docsearch, but don't we expose the templates to the users? If so this would be a breaking change

I'm not so sure, judging from the package exports and the official documentation. The only thing that seems to be exposed is the `docsearch()` factory

Haroenv:
One thing however that was just pointed out to me by @francoischalifour is that hogan deals with XSS, while now the html isn't sanitised anymore, we need to filter out `<` and similar characters: 

& becomes &amp;
< becomes &lt;
> becomes &gt;

not sure if there's attributes with user input that needs to be escaped, but 

" becomes &quot;
' becomes &#39;

source I checked to be sure: https://stackoverflow.com/a/7382028/3185307

francoischalifour:
We're done a complete rewrite for DocSearch.js v3 that gets rid of Hogan.js. This POC looks promising and we might move forward with it in the coming months.

For this reason and the fact that Hogan.js does more work than basic strings (e.g. escaping – which I assume is the reason we went for Hogan.js in the first place), I don't think we're going to move forward with this PR. This is a change that is not radical enough to be included in the next minor version given the risks it generates.

Thank you for the proposition, it's very appreciated! Feel free to open an issue first next time so that we discuss what changes might make it to a release.

PR #796, opened 2019-08-21
doc(wording): remove crawler word cc #774

s-pace:


s-pace:
Please ignore circleci since it is only required for the v3 cc @eunjae-lee

docs/config.json
```diff
@@ -163,6 +163,10 @@
     {
       "from": "recommended-configuration.html",
       "to": "required-configuration.html"
+    },
+    {
+      "from": "crawler-overview.html.html",
```
francoischalifour: There are 2 `.html` extensions.

s-pace: My bad, Solved.

PR #816, opened 2019-09-01
Fix wrong menu link

### LINKED ISSUE #814: Wrong link to scraper overview in navigation
The link to technical overview points to scraper-overview.html which does not exist.

s-pace:
resolves #814

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -74,7 +74,7 @@
       "pages": [
         {
           "title": "DocSearch Technical Overview (the Scraper)",
-          "url": "scraper-overview.html"
+          "url": "crawler-overview.html"
         },
         {
           "title": "Inside The Engine",


docs/config.json
```diff
@@ -74,7 +74,7 @@
       "pages": [
         {
           "title": "DocSearch Technical Overview (the Scraper)",
-          "url": "scraper-overview.html"
+          "url": "crawler-overview.html"
```
Haroenv: don't you want to change the page's name instead?

s-pace: I think it should be enough

Haroenv: i mean changing the file instead of the url in the config :)

s-pace: Let’s keep it this way

PR #819, opened 2019-09-06
doc: feat: add form page

s-pace:
WIP

Needs to avoid redirection and better styling


s-pace:
Needs to avoid CORS or find another way to post the form, integrate into navigation and better styling

s-pace:
Since the "join" endpoint replies with a header: `access-control-allow-origin` set to `https://community.algolia.com` this PR needs to be hosted on https://community.algolia.com to be fully testable even if it is a simple request from a CORS point of view.

@julienbourdeau will broaden the allowed origins. Thanks! Please let us know in this thread.

This PR is ready for merge once we can preview it on this domain or once the allowed origins is broaden.

PR #829, opened 2019-09-11
Docusaurus: remove npm authentication by token

s-pace:


Shipow:
Won't be able to deploy on Netlify without this env as it needs to download the ui-library (not public yet) from our private repo

PR #853, opened 2019-09-27
Fix circle.io logo caption to CircleCI

aaronstillwell:
**Summary**

The CircleCI logo is incorrectly labelled as "circle.io" - this PR changes the `alt` attribute and caption to `CircleCI`, although leaves the image name as-is.

**Result**

Glide now reads "CircleCI" as it should


s-pace:
Thanks @aaronstillwell

PR #857, opened 2019-10-01
Fix typo

jamesrhea:
Change "every internal links" to "every internal link".

**Summary**

Fix minor typo.


s-pace:
Thank you @jamesrhea. Always struggling with it every time_ ;)

PR #884, opened 2019-12-09
Add Orchid to supported integrations

cjbrooks12:
**Summary**

I had reached out to the DocSearch support team a month ago about adding Orchid as an official integration once I released the feature. Now that this is released, I am updating the page here with the link to Orchid's DocSearch integration docs.


s-pace:
Thanks for the great job here @cjbrooks12

PR #885, opened 2019-12-19
Remove duplicate properties

MartijnCuppens:
Minor tweaks: remove duplicate properties


s-pace:
Thanks for the fix @MartijnCuppens

PR #886, opened 2019-12-24
Fix HTML element syntax in docs

rnevius:
**Summary**

As above. [Best practice](https://developer.mozilla.org/en-US/docs/Web/HTML/Element) is to portray HTML tags without a forward slash `/`,
especially when the tags aren't self-closing.


s-pace:
Thanks for your contribution @rnevius

PR #955, opened 2020-05-28
Include the result name in the aria label

orta:
**Summary**

When you use a screen reader with the search box, you only hear "Link to the result" for each result - which means you don't know what the result is. This adds the text of the result to the label so that folks can hear it.

### DIFF
--- a/src/lib/templates.js
+++ b/src/lib/templates.js
@@ -10,7 +10,7 @@ const templates = {
     {{#isCategoryHeader}}${suggestionPrefix}__main{{/isCategoryHeader}}
     {{#isSubCategoryHeader}}${suggestionPrefix}__secondary{{/isSubCategoryHeader}}
     "
-    aria-label="Link to the result"
+    aria-label="Link to the result: {{{subcategory}}}"
     href="{{{url}}}"
     >
     <div class="${suggestionPrefix}--category-header">


orta:
That CI fail doesn't _look_ like my fault 👍

Haroenv:
This looks like an interesting fix, thanks! We are just on the doorstep of releasing the next version of docsearch. The code for that is available here https://github.com/algolia/docsearch/tree/next and @francoischalifour will be available to help you with setting up

francoischalifour:
The new code is actually on [this repo](https://github.com/francoischalifour/autocomplete.js/tree/b3d4e17c5edf89a442a5bc70a707492f895f79cb/packages/docsearch-react) for now. You can learn more about this version in facebook/docusaurus#2815.

We didn't yet do an accessibility audit but I believe this fix also needs to be done for the next version. Would you be interested in porting this fix there @orta?

orta:
I've given it a shot and https://github.com/francoischalifour/autocomplete.js/pull/51 is a result of updating

orta:
While the PR above seems blocked  and on the off-chance someone else hits the same issue, I ended up instead removing the aria label completely (which means the screen reader will use the result text anyway) and then have a few other quick tweaks to make it work better in the different keyboard mode with the screen reader active. 

You can find the most up to date version of that [script here](https://github.com/microsoft/TypeScript-Website/blob/v2/packages/typescriptlang-org/scripts/downloadSearchAssets.js) ([commit locked](https://github.com/microsoft/TypeScript-Website/blob/0885edea5ecb8ec3e337e5781c383c1337bdfd70/packages/typescriptlang-org/scripts/downloadSearchAssets.js))

PR #983, opened 2020-11-04
feat: push keyboard buttons till client

maraisr:
This PR aims to solve the issue around flickering buttons when we hydrate not he client and figure out we're not on Mac.

This PR also aims to make the build flow compatible for windows developers, removing the `rm -rf` and environment things to use npm proxy scripts, both `cross-env` and `rimraf` are 2 very popular libraries for this.

fixes: https://github.com/facebook/docusaurus/issues/3682


francoischalifour (approved):
Thank you for taking the time to tackle this 👍

PR #989, opened 2020-11-26
fix(website): add workspace `nohoist` for Docusaurus

slorber:
cc @shipow @francoischalifour


slorber:
Still having build errors, but it's not related to the same problem afaik

image

Shipow:
it solved the problem, thanks @slorber ;)

francoischalifour (approved):
Thank you!

PR #991, opened 2020-12-07
fix(css): remove `sideEffects` package field

chenxsan:
**Summary**

When using `@docsearch/css` with webpack like below:

```
import '@docsearch/css'
```
[`sideEffects: false`](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) would cause webpack to prune all the css.

I have to work around with a hack like this:

```
import styles from '@docsearch/css';
// do something with `styles` to keep it around.
```

**Result**

### DIFF
--- a/packages/docsearch-react/package.json
+++ b/packages/docsearch-react/package.json
@@ -9,7 +9,6 @@
     "name": "Algolia, Inc.",
     "url": "https://www.algolia.com"
   },
-  "sideEffects": false,
   "files": [
     "dist/",
     "style/",
--- a/packages/docsearch-react/package.json
+++ b/packages/docsearch-react/package.json
@@ -9,6 +9,7 @@
     "name": "Algolia, Inc.",
     "url": "https://www.algolia.com"
   },
+  "sideEffects": false,
   "files": [
     "dist/",
     "style/",
--- a/packages/docsearch-css/package.json
+++ b/packages/docsearch-css/package.json
@@ -9,7 +9,6 @@
     "name": "Algolia, Inc.",
     "url": "https://www.algolia.com"
   },
-  "sideEffects": false,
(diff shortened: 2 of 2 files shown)


packages/docsearch-react/package.json
```diff
@@ -9,7 +9,6 @@
     "name": "Algolia, Inc.",
     "url": "https://www.algolia.com"
   },
-  "sideEffects": false,
```
Haroenv, line 12: it should be this line, not the one you changed, this is the package.json for the react package, not the css package

https://github.com/algolia/docsearch/blob/e68441b51cecf5170e9d01de37baa1bd25956378/packages/docsearch-css/package.json#L12

Haroenv (commented):
this change makes sense according to me, although I haven't yet checked it. The change should happen in a different file though

packages/docsearch-react/package.json
chenxsan: But I was using `@docsearch/css` package in my case. There's no problem with `@docsearch/react` package even with `sideEffects: false` presented in its `package.json`. Do you want me to set up a repository so you can check?

Haroenv: the change in this PR was made to the react package, instead of the css package, you might have confused the two?

chenxsan: > the change in this PR was made to the react package, instead of the css package, you might have confused the two?

Oops, sorry, it's my fault.

chenxsan: Updated, thanks for heads-up.

Haroenv (approved):
I think this is correct!

francoischalifour (approved):
Thanks!

PR #993, opened 2021-02-01
fix: improve contrast ratio for button placeholder

lex111:
**Summary**

Currently there is a11y error in light mode related to insufficient color contrast in the placeholder button.

image

**Result**

Checker: https://webaim.org/resources/contrastchecker/

| Before   | After    |
| -------- | -------- |
|  |  |

cc @francoischalifour


lex111:
Just friendly reminder, since I would like to see this fix in new version.

Shipow:
Hi @lex111 , we've acknowledged the problem but the fix seams to break some important hierarchy info so we need to rework on this.
Thanks for pointing this. When do you plan to release a new version?

lex111:
I don't know exactly, but I wanted to do it as soon as possible. So I will wait for your action.

PR #994, opened 2021-02-01
fix: use spans inside button tag instead of divs

lex111:
**Summary**

At the moment there is the following HTML issue related to the incorrect use of the `button` tag:

```
Element “div” not allowed as child of element “button” in this context. (Suppressing further errors from this subtree.)
```

I suggest replacing the `button` with a regular `div` element.

**Result**

According to the validator (https://validator.w3.org/nu/#textarea), the error has been fixed.

cc @francoischalifour

### DIFF
--- a/packages/docsearch-react/src/DocSearchButton.tsx
+++ b/packages/docsearch-react/src/DocSearchButton.tsx
@@ -30,20 +30,26 @@ export const DocSearchButton = React.forwardRef<
   }, []);
 
   return (
-    <div className="DocSearch DocSearch-Button" {...props} ref={ref}>
-      <div className="DocSearch-Button-Container">
+    <button
+      type="button"
+      className="DocSearch DocSearch-Button"
+      aria-label="Search"
+      {...props}
+      ref={ref}
+    >
+      <span className="DocSearch-Button-Container">
         <SearchIcon />
         <span className="DocSearch-Button-Placeholder">Search</span>
-      </div>
+      </span>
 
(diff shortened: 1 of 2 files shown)


packages/docsearch-react/src/DocSearchButton.tsx
```diff
@@ -30,13 +30,7 @@ export const DocSearchButton = React.forwardRef<
...
   return (
-    <button
-      type="button"
-      className="DocSearch DocSearch-Button"
-      aria-label="Search"
-      {...props}
-      ref={ref}
-    >
+    <div className="DocSearch DocSearch-Button" {...props} ref={ref}>
       <div className="DocSearch-Button-Container">
```
Haroenv, line 34: this should be moved to a span instead and the button stays (I think)

Shipow: Yes, we are more and more confident with this button tag.

lex111: Agree, it makes sense, replaced divs with spans.

Shipow (approved):
Thanks a lot Alexey and Haroen!

Haroenv (approved):
this looks like a solid fix, thanks!

francoischalifour (approved):
Thank you @lex111!

francoischalifour:
Released in `@docsearch/react@3.0.0-alpha.33`.

PR #995, opened 2021-02-03
feat: add `translations` prop to search button

slorber:
**Summary**

Ability to translate DocSearch placeholder label, a low hanging fruit as it's the most visible one, but there are probably other labels to allow translating (including aria attributes).

This is a pseudo code suggestion, not sure it works yet :) 

You may be happy to know about `ComponentProps` in TS :) 

**Result**

image

### DIFF
--- a/packages/docsearch-react/src/DocSearchButton.tsx
+++ b/packages/docsearch-react/src/DocSearchButton.tsx
@@ -3,10 +3,7 @@ import React, { useEffect, useState } from 'react';
 import { ControlKeyIcon } from './icons/ControlKeyIcon';
 import { SearchIcon } from './icons/SearchIcon';
 
-export type DocSearchButtonProps = React.DetailedHTMLProps<
-  React.ButtonHTMLAttributes<HTMLButtonElement>,
-  HTMLButtonElement
->;
+export type DocSearchButtonProps = React.ComponentProps<'button'> & {placeholderLabel?: string}
 
 const ACTION_KEY_DEFAULT = 'Ctrl' as const;
 const ACTION_KEY_APPLE = '⌘' as const;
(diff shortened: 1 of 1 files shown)


lex111 (commented):
@francoischalifour please take a look at this.

francoischalifour (commented):
Thank you @slorber!

There will be other texts that we want to translate like the `aria-label` of the button. So I think we should rather pass a `translations` prop with a few keys:

```ts
type Translations = Partial<{
  buttonText: string;
  buttonAriaLabel: string;
}>
```

Wdyt?

(Thanks for letting us know about `React.ComponentProps`.)

slorber:
sure, just updated

francoischalifour (approved):
Thanks!

Haroenv (approved):

francoischalifour:
Released in `@docsearch/react@3.0.0-alpha.33`.

slorber:
thanks :)

meteorlxy:
Seems that this prop is not available in the `<DocSearch>` component?

francoischalifour:
@meteorlxy This prop is only available on the `DocSearchButton` component, exported from the package. It's not yet on the `DocSearch` component itself. We plan to add it in the future.

slorber:
@francoischalifour was wondering if you could create an issue to track i18n support?

Some Docusaurus users asked about this and want to have a link to give them ;)

https://github.com/facebook/docusaurus/issues/3643#issuecomment-833649796

shortcuts:
Hi @slorber,

Here's some resources you can share with your users:

As DocSearch v3 rely on [Autocomplete v1](https://github.com/algolia/autocomplete) ([which has just been released](https://discourse.algolia.com/t/autocomplete-v1-is-released/12549)), we are re-thinking the way we would like to ship a stable version of it.

The next steps are:
- Implement the `translations API` in Autocomplete
- Package DocSearch v3 directly with Autocomplete (which would offer more flexibility/features and in this case, the `translations API`)

We will start working on it in the coming months, and will make sure to let you know what's the progress on it!

slorber:
thanks :)

stnor:
@shortcuts Any progress on this yet? Estimated arrival? Thanks.

PR #999, opened 2021-03-05
fix(deps): Update and pin Autocomplete to the latest version

shortcuts:
**Summary**

We introduced the required `sourceId` prop in https://github.com/algolia/autocomplete/pull/429 and _I assume that_ as the Autocomplete version wasn't pinned, the latest DocSearch version throw an error.

This implementation provides a default `sourceId` to the DocSearch sources.

_Other improvements might be needed but this should fix the current issue._

### DIFF
--- a/packages/docsearch-react/src/DocSearchModal.tsx
+++ b/packages/docsearch-react/src/DocSearchModal.tsx
@@ -51,13 +51,13 @@ export function DocSearchModal({
     completion: null,
     context: {},
     isOpen: false,
-    selectedItemId: null,
+    activeItemId: null,
     status: 'idle',
   });
 
   const containerRef = React.useRef<HTMLDivElement | null>(null);
   const modalRef = React.useRef<HTMLDivElement | null>(null);
-  const searchBoxRef = React.useRef<HTMLDivElement | null>(null);
+  const formElementRef = React.useRef<HTMLDivElement | null>(null);
   const dropdownRef = React.useRef<HTMLDivElement | null>(null);
   const inputRef = React.useRef<HTMLInputElement | null>(null);
   const snippetLength = React.useRef<number>(10);
@@ -139,6 +139,7 @@ export function DocSearchModal({
 
             return [
               {
+                sourceId: 'DocSearchRecentSearches',
                 onSelect({ item, event }) {
                   saveRecentSearch(item);
 
@@ -154,6 +155,7 @@ export function DocSearchModal({
                 },
               },
               {
+                sourceId: 'DocSearchFavoriteSearches',
                 onSelect({ item, event }) {
                   saveRecentSearch(item);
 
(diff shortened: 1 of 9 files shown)


packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -139,6 +139,7 @@ export function DocSearchModal({
 
             return [
               {
+                sourceId: 'DocSearchRecentSearches',
```
francoischalifour, line 142: From algolia/autocomplete#490 we'll recommend naming those as `recentSearches` to pick the `sourceId`s easily with destructuring. We can start naming those following this convention already.

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -239,6 +241,7 @@ export function DocSearchModal({
 
               return Object.values<DocSearchHit[]>(sources).map((items) => {
                 return {
+                  sourceId: 'DocSearchHits',
```
francoischalifour, line 244: All the DocSearch section will get the same `sourceId`s here. Can we differentiate them?

francoischalifour (commented):
Thanks!

- We should be safe removing the [`enterKeyHint` line](https://github.com/algolia/docsearch/blob/30a7a229ff9490d2fea567b92c1404489f90cc43/packages/docsearch-react/src/SearchBox.tsx#L65) which is [now supported in Autocomplete alpha 38](https://github.com/algolia/autocomplete/blob/next/CHANGELOG.md#features-6).
- We need to update `defaultSelectedItemId` to `defaultActiveItemId`, otherwise the first result is not automatically active. (Why didn't we get a TypeScript error?)

packages/docsearch-react/src/DocSearchModal.tsx
shortcuts: Make sense!

packages/docsearch-react/src/DocSearchModal.tsx
shortcuts: Fixed in https://github.com/algolia/docsearch/pull/999/commits/fc965daeb4877a3febd619012dff39febc7f5de7

packages/docsearch-react/src/DocSearchModal.tsx
shortcuts: Fixed in https://github.com/algolia/docsearch/pull/999/commits/fc965daeb4877a3febd619012dff39febc7f5de7

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -237,43 +239,46 @@ export function DocSearchModal({
...
-                                  siblingItem.type === 'lvl1' &&
-                                  siblingItem.hierarchy.lvl1 ===
-                                    item.hierarchy.lvl1
-                              ),
-                          };
-                        })
+              return Object.values<DocSearchHit[]>(sources).map(
+                (items, index) => {
+                  return {
+                    sourceId: `DocSearch-Hits-${index}`,
```
francoischalifour, line 245:
```suggestion
                    sourceId: `hits${index}`,
```

francoischalifour (approved):

PR #1001, opened 2021-03-31
fix(results): unescaped lvl0

shortcuts:
**Summary**

In some cases, the hits returned by the engine can contain escaped characters. We now retrieve the `lvl0` from the `_highlightResults` (which are always unescaped) and remove the highlighting tags.

| before | after |
| --- | --- |
| Screenshot 2021-03-31 at 16 55 12 |  Screenshot 2021-03-31 at 16 59 27|

### DIFF
--- a/packages/docsearch-react/src/ResultsScreen.tsx
+++ b/packages/docsearch-react/src/ResultsScreen.tsx
@@ -17,7 +17,7 @@ export function ResultsScreen(props: ResultsScreenProps) {
         }
 
         const title = removeHighlightTags(
-          collection.items[0]._highlightResult.hierarchy_camel[0].lvl0.value
+          collection.items[0]._highlightResult.hierarchy.lvl0.value
         );
 
         return (
--- a/packages/docsearch-react/src/ResultsScreen.tsx
+++ b/packages/docsearch-react/src/ResultsScreen.tsx
@@ -4,7 +4,7 @@ import { SelectIcon, SourceIcon } from './icons';
 import { Results } from './Results';
 import { ScreenStateProps } from './ScreenState';
 import { InternalDocSearchHit } from './types';
-import { unescape } from './utils';
+import { removeHighlightTags } from './utils';
 
 type ResultsScreenProps = ScreenStateProps<InternalDocSearchHit>;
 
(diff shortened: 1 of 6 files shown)


packages/docsearch-react/src/utils/unescape.ts
```diff
@@ -0,0 +1,8 @@
+const regexEscapedHtml = /(<mark>|<\/mark>)/g;
+const regexHasEscapedHtml = RegExp(regexEscapedHtml.source);
+
+export function unescape(value: string): string {
```
shortcuts, line 4: Just realized that `unescape` can be confusing, maybe `removeHighlightTags`?

Shipow (approved):

packages/docsearch-react/src/ResultsScreen.tsx
```diff
@@ -15,7 +16,9 @@ export function ResultsScreen(props: ResultsScreenProps) {
           return null;
         }
 
-        const title = collection.items[0].hierarchy.lvl0;
+        const title = unescape(
+          collection.items[0]._highlightResult.hierarchy_camel[0].lvl0.value
```
francoischalifour, line 20: Why do we use the camel version?

shortcuts: I matched the v2 implementation, is there any differences/preferences expect that `camel` is an array?

packages/docsearch-react/src/utils/unescape.ts
shortcuts: renamed in https://github.com/algolia/docsearch/pull/1001/commits/b56836259d96aeabdcc4a46a59bb3ea9030daef9

packages/docsearch-react/src/ResultsScreen.tsx
shortcuts: replaced in https://github.com/algolia/docsearch/pull/1001/commits/5ea358bd4b441c74ae83a1573ba9ab81370d7261 as we discussed with @Shipow

francoischalifour (approved):

ericxjohnston:
@shortcuts Thanks Clement for merging these changes! Do you know when the next alpha version will be released?

shortcuts:
Hi @ericxjohnston, the version `v3.0.0-alpha.35` is now released.

ericxjohnston:
@shortcuts Thanks Clement!

Issue #1018, opened 2021-07-15
Send search request to Algolia with some delay


Markeli:
## Describe the problem

Now a new search request is sending to Aloglia on each query change. I want to reduce the number of requests to Aloglia. It would be nice if Aloglia client will have some delay option in config. If the delay is set, the request will be sent if the user is not printing. Without delay in the config, the search behavior will be the same as now.

## Describe the solution

Add new option to Algolia search plugin config. If the delay is set search query will be sent only after a user stops inputting the query. I assume that desired behavior can be easily achieved by React hooks. There is an example on StackOverflow: https://stackoverflow.com/questions/53071774/reactjs-delay-onchange-while-typing

Haroenv:
Could you explain why you want to limit the number of requests?

Markeli:
There are two reasons:
- to improve the search experience for users with a slow Internet connection
- to reduce costs of Algolia search. 

For example, a user wants to find articles with the word `project`. With the current solution, there are will be 7 separate search requests (for `p`/`pr`/`pro`/`proj`/`proje`/`projec`/`project`).

Markeli:
In other words, we want to disable something like `searchAsYouType` option. 

By the way, this option has already existed in `instantSeacrh.js`.
![image](https://user-images.githubusercontent.com/5919274/125783042-9ae42cb7-c58a-44f6-8fca-81b554816ace.png)

Haroenv:
Because docsearch is mainly meant for implementations for which we cover the cost, there so far hasn't been a reason to debounce the search

Haroenv:
With the PR that was just merged (once that's released), you can debounce like this:


```ts
type Func = (...args: any[]) => any;

type DebouncedFunction<TFunction extends Func> = (
  this: ThisParameterType<TFunction>,
  ...args: Parameters<TFunction>
) => Promise<ReturnType<TFunction>>;

// Debounce a function call to the trailing edge.
// The debounced function returns a promise.
export function debounce<TFunction extends Func>(
  func: TFunction,
  wait: number
): DebouncedFunction<TFunction> {
  let lastTimeout: ReturnType<typeof setTimeout> | null = null;
  return function (...args) {
    // @ts-ignore-next-line
    const that = this;
    return new Promise((resolve, reject) => {
      if (lastTimeout) {
        clearTimeout(lastTimeout);
      }
      lastTimeout = setTimeout(() => {
        lastTimeout = null;
[...]

PR #1019, opened 2021-07-15
refactor(searchClient): use client `search` method

shortcuts:
**Summary**

We now use the multi-queries `search` method from the client which allows the user to override the method from `transformSearchClient`.

**Result**

The user can now override the `search` method, to debounce the search for example.

**Example**

[sandbox](https://codesandbox.io/s/docsearch-v3-debounced-search-gnx87)

**Issues**

- Fixes https://github.com/algolia/docsearch/issues/1018
- https://github.com/facebook/docusaurus/issues/5174

### DIFF
--- a/packages/docsearch-react/src/DocSearchModal.tsx
+++ b/packages/docsearch-react/src/DocSearchModal.tsx
@@ -173,35 +173,40 @@ export function DocSearchModal({
           }
 
           return searchClient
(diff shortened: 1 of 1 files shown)


Haroenv (approved):
it's a refactor, not a fix

francoischalifour (approved):

PR #1021, opened 2021-07-16
fix: eliminate keyboard shortcuts UI flash

chenxsan:
There's a flash of unstyled content regarding the search box:

https://user-images.githubusercontent.com/1091472/125940962-202531fe-58aa-40f5-af9c-fc775eb7c9bf.mov

You can check it on https://docsearch.algolia.com/

Setting a width would eliminate the problem, you can check it on https://webpack.js.org/ for how it looks.

### DIFF
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -12,7 +12,6 @@
   margin: 0 0 0 16px;
   padding: 0 8px;
   user-select: none;
-  min-width: 155px;
 }
 
 .DocSearch-Button:hover,
@@ -63,10 +62,6 @@
 }
 
 @media (max-width: 750px) {
-  .DocSearch-Button {
-    min-width: auto;
-  }
-
   .DocSearch-Button-KeySeparator,
   .DocSearch-Button-Key {
     display: none;
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -12,7 +12,7 @@
   margin: 0 0 0 16px;
   padding: 0 8px;
   user-select: none;
-  width: 155px;
+  min-width: 155px;
 }
 
 .DocSearch-Button:hover,
@@ -64,7 +64,7 @@
 
 @media (max-width: 750px) {
   .DocSearch-Button {
-    width: auto;
+    min-width: auto;
   }
 
   .DocSearch-Button-KeySeparator,
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -12,6 +12,7 @@
   margin: 0 0 0 16px;
   padding: 0 8px;
   user-select: none;
+  width: 155px;
 }
 
 .DocSearch-Button:hover,
@@ -62,6 +63,10 @@
 }
 
 @media (max-width: 750px) {
+  .DocSearch-Button {
+    width: auto;
+  }
+
   .DocSearch-Button-KeySeparator,
   .DocSearch-Button-Key {
     display: none;
--- a/packages/docsearch-react/src/DocSearchButton.tsx
+++ b/packages/docsearch-react/src/DocSearchButton.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 
 import { ControlKeyIcon } from './icons/ControlKeyIcon';
 import { SearchIcon } from './icons/SearchIcon';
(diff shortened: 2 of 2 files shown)


packages/docsearch-css/src/button.css
```diff
@@ -12,6 +12,7 @@
   margin: 0 0 0 16px;
   padding: 0 8px;
   user-select: none;
+  width: 155px;
```
shortcuts, line 15:
```suggestion
  min-width: 155px;
```

packages/docsearch-css/src/button.css
```diff
@@ -62,6 +63,10 @@
 }
 
 @media (max-width: 750px) {
+  .DocSearch-Button {
+    width: auto;
```
shortcuts, line 67:
```suggestion
    min-width: auto;
```

shortcuts (commented):
Thanks for your PR, It indeed looks better! Small comment, I think using `min-width` might be safer

packages/docsearch-css/src/button.css
```diff
@@ -12,6 +12,7 @@
   margin: 0 0 0 16px;
   padding: 0 8px;
   user-select: none;
+  min-width: 155px;
```
Haroenv, line 15: how do we know the value is 155px?

chenxsan: I just inspected the element for the value:

CleanShot 2021-07-16 at 21 08 30@2x

Not a exact one, but it works in the webpack site case.

shortcuts: The computed size of the component loaded is `155.359px`, I assume that's where @chenxsan also found it.

Haroenv: What if a different font, placeholder or zoom level is used? I guess it's still a safe default, but I think in some cases it will still flash

chenxsan: I think users could always customize it to their own use cases. We're just providing a default one which should hopefully eliminate the flash for most of the cases. For instance, the value would work for https://docusaurus.io/ as well.

shortcuts: > What if a different font, placeholder or zoom level is used?

Most of these won't have effect unless the user is already overriding the provided styles. I think this could be a safe default and it is requested quite frequently.

wdyt? @Haroenv

Haroenv (dismissed):
overall we rely on px too much, but this is a sensible default

packages/docsearch-css/src/button.css
francoischalifour: It's safer with the `min-` change IMO.

But I agree with Haroen. I feel like setting the constraint on `.DocSearch-Button-Keys` is better because that's ultimately what we want as constant.

chenxsan: @francoischalifour I don't think it would work on `.DocSearch-Button-Keys` considering it's controlled by a state value https://github.com/algolia/docsearch/blob/1ee5354ea5026c531eb33a37d5eca8b66b7eca9d/packages/docsearch-react/src/DocSearchButton.tsx#L51

francoischalifour: Ah right, we can still move that check one line below to always render that element though?

chenxsan: Thanks for bringing up this as I found a problem with my solution.

There're two cases depending on the `key` state:

1. a search box with the keys
CleanShot 2021-07-21 at 16 05 27@2x

2. a search box without the keys
CleanShot 2021-07-21 at 16 05 07@2x


When there's no keys, the current solution would be deficient, as it results in a wide search box:
CleanShot 2021-07-21 at 16 10 17@2x

I'll see what I can do with it.

chenxsan:
I've refactor the code as the former proposal won't cover all use cases, see  https://github.com/algolia/docsearch/pull/1021#discussion_r673894680 for detail.


https://user-images.githubusercontent.com/1091472/126486014-e2166aaa-d128-438b-a976-c5860586982d.mov

Let me know what you think about the new code.

francoischalifour:
@chenxsan Unfortunately we cannot change the `DocSearchButton` API since it's public API (and therefore will break the Docusaurus implementation). However we can still use your strategy but move the lazy state evaluation in the `DocSearchButton` component.

I'm not sure how the last change fixes the issue since you didn't change the CSS, could it be that you didn't re-build the CSS?

chenxsan:
@francoischalifour You are right! I didn't notice it's public API. I've updated the code.

packages/docsearch-react/src/DocSearchButton.tsx
```diff
@@ -25,15 +25,14 @@ export const DocSearchButton = React.forwardRef<
 >(({ translations = {}, ...props }, ref) => {
   const { buttonText = 'Search', buttonAriaLabel = 'Search' } = translations;
 
-  const [key, setKey] = useState<
+  const [key] = useState<
```
francoischalifour, line 28: This can be a `useMemo` now that we don't update the `key` value 🙂

chenxsan: Sure, updated.

francoischalifour (approved):
Much simpler now, thank you @chenxsan!

chenxsan:
@francoischalifour @shortcuts  Unfortunately it doesn't fix the flashing problem. I've filed a pull request on webpack documentation here https://github.com/webpack/webpack.js.org/pull/5224, and here's the preview url https://webpack-js-org-git-fork-chenxsan-upgrade-do-a4186d-webpack-docs.vercel.app/, you can compare it with https://webpack.js.org/ which has the width set for the search box.

PR #1033, opened 2021-07-23
fix(button): prevent container keys ui flash

shortcuts:
_Follow up of https://github.com/algolia/docsearch/pull/1021_

**Summary**

This PR prevent the UI flash for server-side rendered websites. Since [the `key` value](https://github.com/algolia/docsearch/blob/f38276308a4e0ff140e4c1d4db1beced4f610068/packages/docsearch-react/src/DocSearchButton.tsx#L31-L33) is `null`, the `keys` container isn't rendered and therefore causes a UI flash.

**Result**

The condition is now a level below (on the `key`s), and we've added a fixed `width` to the container.

**Preview**

https://user-images.githubusercontent.com/20689156/126791441-7ef26e5f-3b5d-4002-aee3-9fb627634680.mov

**Related**
- https://github.com/webpack/webpack.js.org/pull/5224
- https://github.com/facebook/docusaurus/pull/5214

### DIFF
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -63,11 +63,8 @@
 }
 
 @media (max-width: 750px) {
-  .DocSearch-Button-KeySeparator,
-  .DocSearch-Button-Key {
-    display: none;
-  }
-
+  .DocSearch-Button-Keys,
+  .DocSearch-Button-Key,
   .DocSearch-Button-Placeholder {
     display: none;
   }
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -43,6 +43,7 @@
 
 .DocSearch-Button-Keys {
   display: flex;
+  width: calc(2 * 20px + 2 * 0.4em);
 }
 
 .DocSearch-Button-Key {
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -43,7 +43,7 @@
 
 .DocSearch-Button-Keys {
   display: flex;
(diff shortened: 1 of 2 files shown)


packages/docsearch-css/src/button.css
```diff
@@ -62,11 +63,8 @@
 }
 
 @media (max-width: 750px) {
-  .DocSearch-Button-KeySeparator,
-  .DocSearch-Button-Key {
-    display: none;
-  }
-
+  .DocSearch-Button-Keys,
+  .DocSearch-Button-Key,
```
francoischalifour, line 67: That line shouldn't be necessary since we hide its container.

packages/docsearch-css/src/button.css
```diff
@@ -43,6 +43,7 @@
 
 .DocSearch-Button-Keys {
   display: flex;
+  width: calc(2 * 20px + 2 * 0.4em);
```
francoischalifour, line 46: I think we should keep it loose for people using bigger fonts for instance.

```suggestion
  min-width: calc(2 * 20px + 2 * 0.4em);
```

packages/docsearch-css/src/button.css
shortcuts: fixed in https://github.com/algolia/docsearch/pull/1033/commits/bfa436a97e03f26310e9989847e1792c5f5a61ef

packages/docsearch-css/src/button.css
shortcuts: Indeed, updated in https://github.com/algolia/docsearch/pull/1033/commits/bfa436a97e03f26310e9989847e1792c5f5a61ef

francoischalifour (approved):

PR #1039, opened 2021-08-19
fix(removeHighlightTags): prevent destructuring `undefined` object

shortcuts:
**Summary**

In case the indexing isn't optimal, an error is thrown as we are destructuring a possibly null `lvl0`.

**Reproduction**
This can happen when using DocSearch with a Netlify x Algolia Crawler index, but also with a DocSearch index: Search for `ge` in [this sandbox](https://codesandbox.io/s/trusting-elion-8479m?file=/src/index.js:78-198)

**Solution**

We now use `getAttributeValueByPath` to retrieve the attribute `value`, and fallback to an empty string if not found (alternatively, we could provide a default value such as `Documentation`).

### DIFF
--- a/packages/docsearch-react/src/utils/getAttributeValueByPath.ts
+++ b/packages/docsearch-react/src/utils/getAttributeValueByPath.ts
@@ -1,6 +1,0 @@
-export function getAttributeValueByPath<TRecord>(
-  record: TRecord,
-  path: string[]
-): any {
-  return path.reduce((current, key) => current && current[key], record);
-}
--- a/packages/docsearch-react/src/utils/getAttributeValueByPath.ts
+++ b/packages/docsearch-react/src/utils/getAttributeValueByPath.ts
@@ -1,0 +1,6 @@
+export function getAttributeValueByPath<TRecord>(
+  record: TRecord,
+  path: string[]
+): any {
+  return path.reduce((current, key) => current && current[key], record);
+}
--- a/packages/docsearch-react/src/utils/removeHighlightTags.ts
+++ b/packages/docsearch-react/src/utils/removeHighlightTags.ts
@@ -1,5 +1,4 @@
 import { DocSearchHit, InternalDocSearchHit } from './../types';
-import { getAttributeValueByPath } from './getAttributeValueByPath';
 
 const regexHighlightTags = /(<mark>|<\/mark>)/g;
 const regexHasHighlightTags = RegExp(regexHighlightTags.source);
@@ -13,11 +12,11 @@ export function removeHighlightTags(
     return hit.hierarchy.lvl0;
   }
 
-  const value: string =
-    getAttributeValueByPath(
-      hit._highlightResult ? hit : internalDocSearchHit.__docsearch_parent,
-      ['_highlightResult', 'hierarchy', 'lvl0', 'value']
-    ) || '';
+  const { value } =
+    (internalDocSearchHit.__docsearch_parent
+      ? internalDocSearchHit.__docsearch_parent?._highlightResult?.hierarchy
+          ?.lvl0
+      : hit._highlightResult?.hierarchy?.lvl0) || '';
 
   return value && regexHasHighlightTags.test(value)
     ? value.replace(regexHighlightTags, '')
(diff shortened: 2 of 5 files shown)


packages/docsearch-react/src/utils/removeHighlightTags.ts
```diff
@@ -1,22 +1,23 @@
...
   }
 
-  const { value } = hit._highlightResult
-    ? hit._highlightResult.hierarchy.lvl0
-    : (hit as InternalDocSearchHit).__docsearch_parent!._highlightResult
-        .hierarchy.lvl0;
+  const value: string =
+    getAttributeValueByPath(
+      hit._highlightResult ? hit : internalDocSearchHit.__docsearch_parent,
+      ['_highlightResult', 'hierarchy', 'lvl0', 'value']
```
Haroenv, line 19: could this be done with ?. as well to keep type safety?

packages/docsearch-react/src/utils/getAttributeValueByPath.ts
```diff
@@ -0,0 +1,6 @@
+export function getAttributeValueByPath<TRecord>(
+  record: TRecord,
+  path: string[]
+): any {
```
Haroenv, line 4: you actually can type this if path is a readonly array, but not worth the hassle (like https://dev.to/pouja/making-lodash-function-get-type-safe-313o, but that's with a string, although the main part is the same)

packages/docsearch-react/src/utils/removeHighlightTags.ts
shortcuts: I've tried and thought it was a bit messy, so I went with what we use in Autocomplete. Would you prefer the other solution?

Haroenv: isn't the use case in autocomplete with a user-provided path? in this case the path is static, so it's more efficient to do ?.. Why did you find it more messy?

shortcuts: It's indeed not that messier than the other version, is that what you had in mind? https://github.com/algolia/docsearch/pull/1039/commits/2673afcd1aec63021840be1195a8a9d9ae5920d8

packages/docsearch-react/src/utils/removeHighlightTags.ts
```diff
@@ -6,17 +6,17 @@ const regexHasHighlightTags = RegExp(regexHighlightTags.source);
...
 
-  const { value } = hit._highlightResult
-    ? hit._highlightResult.hierarchy.lvl0
-    : (hit as InternalDocSearchHit).__docsearch_parent!._highlightResult
-        .hierarchy.lvl0;
+  const { value } =
+    (internalDocSearchHit.__docsearch_parent
+      ? internalDocSearchHit.__docsearch_parent?._highlightResult?.hierarchy
+          ?.lvl0
+      : hit._highlightResult?.hierarchy?.lvl0) || '';
```
Haroenv, line 19:
```suggestion
      : hit._highlightResult?.hierarchy?.lvl0) || {};
```

you're trying to restructure from '' otherwise right?

Haroenv: are there any tests we can add this case to?

shortcuts: > you're trying to destructure from '' otherwise right?

Both returns `undefined` but yours is more correct indeed, I wonder if we should default it to something like `Documentation` to avoid empty `lvl0` like this:

Screenshot 2021-08-19 at 17 11 46


> are there any tests we can add this case to?

There's no tests other than cypress on this repo yet. Since we're planning to re-write the component after the Reshape API has been introduced, I'm not sure if I should add some now

Haroenv: I'm not sure adding a static item when lvl0 is missing makes sense, the UI would have a lot of useless info

shortcuts: It's not a big deal anyway and I've seen this happen only twice in the past year

Haroenv (approved):

pixelastic:
Thanks @Haroenv and @shortcuts

Issue #1050, opened 2021-09-29
Triggering the search outside of user input


hakankaraduman:
## Description

I want to support duckduckgo !bang feature and this requires me to input the search query with code and trigger the search action with code, without requiring action from the user.

Currently I don't see an api for this, I'm using docsearch.js as described in the documentation.

sarahdayan:
Hey @hakankaraduman!

Can you tell me more on what you're trying to achieve? What would be the consequence of a user using a bang in your search? Do you want to update the search parameters sent to Algolia? Search in another search engine? Trigger custom behaviors unrelated to search?

Depending on what you want to do, the solution will be different. For example, if you're using DocSearch v2, you might be able to perform your custom logic in the [`queryHook`](https://docsearch.algolia.com/docs/behavior#queryhook).

Note that DocSearch.js is meant for an out-of-the-box usage, and consequently doesn't expose too many APIs. If you're trying to do more complex things, we recommend using our [Autocomplete](https://github.com/algolia/autocomplete) library directly (see [docs](http://alg.li/autocomplete)). This will let you do a lot more, as you can [control the sources](https://www.algolia.com/doc/ui-libraries/autocomplete/core-concepts/sources/) based on the query.

To use Autocomplete with your DocSearch account, all you need is to [import and initialize the `algoliasearch` client](https://www.algolia.com/doc/api-client/getting-started/install/javascript/) manually with your DocSearch credentials, then make a search to your DocSearch index as documented [here](https://www.algolia.com/doc/ui-libraries/autocomplete/core-concepts/sources/#using-dynamic-sources). As you can see in the docs, the `getSources` function exposes the `query`, which lets you perform custom logic before you return sources. You can detect a bang usage here, and decide what to do based on its presence.

Let us know how it goes!

hakankaraduman:
Hi Sarah,

What I want is to create an endpoint in my site like `mysite.com/search/?term=mySearchTerm` and give the docsearch library this `mySearchTerm` value that comes from the url and trigger a search as regular. Can I do that without changing the implementation to autocomplete?

hakankaraduman:
Thank you Sarah, v3 solved my issue.

PR #1051, opened 2021-09-30
fix(v3): pass `initialQuery`

shortcuts:
**Summary**

`initialQuery` wasn't passed from the props to the state, which made it unusable.

see [sandbox](https://codesandbox.io/s/hopeful-fast-fjdof?file=/src/App.js)

**Result**

Search input is initialized with `initialQuery`

see [sandbox](https://codesandbox.io/s/festive-sara-7o8nf)

Related issue: https://github.com/algolia/docsearch/issues/1050


bodinsamuel (approved):
LGTM

I see no unit in this repo, we might want to make the stack a bit more reliable at some point.

shortcuts:
> I see no unit test in this repo, we might want to make the stack a bit more reliable at some point.

Since we mostly rely on Autocomplete, it's tested on that side, but we indeed should at least validate props.

I'll add a task to the backlog

PR #1056, opened 2021-10-11
docs(website): add redirects to legacy scraper

shortcuts:
**Summary**

Add Netlify `_redirects` file for the `docsearch-scraper`


shortcuts:
> don't you want a redirect from every old page instead of just the run-your-own? or do the new ones have the same url?

In some cases we do have identical URLs, but I'm indeed listing the ones that we should redirect, I'll push a new commit soon

PR #1198, opened 2021-12-01
docs(links): fix relative path

shortcuts:
## Summary

When navigating on the doc from URLs with trailing slash, links with relative paths were broken.

This PR fixes the relative paths so the navigation works from URL with/without trailing slash.

## Reproduction

- Go to https://docsearch.algolia.com/docs/docsearch-v3/
- Click on the link `custom variables to your config`


Haroenv (approved):

bodinsamuel (approved):
relative URLs are the worst :D

PR #1200, opened 2021-12-03
fix(DocSearchProps): type `searchParameters`

shortcuts:
## Summary

`searchParameters` were not typed


bodinsamuel (approved):
I'll advice to always use strict mode on typescript with no-explicit-any eslint rule

shortcuts:
> I'll advice to always use strict mode on typescript with no-explicit-any eslint rule 

Indeed, I'll add it in a future PR to do a pass on the types, before the stable release

PR #1259, opened 2022-01-12
fix(logo): empty utm campaign

shortcuts:
## Summary

Due to `noreferrer`, the campaign was always empty when clicking the Algolia logo.

We now inline parameters to prevent this issue.

### DIFF
--- a/packages/docsearch-react/src/AlgoliaLogo.tsx
+++ b/packages/docsearch-react/src/AlgoliaLogo.tsx
@@ -10,9 +10,10 @@ type AlgoliaLogoProps = {
 
 export function AlgoliaLogo({ translations = {} }: AlgoliaLogoProps) {
   const { searchByText = 'Search by' } = translations;
+
   return (
     <a
-      href="https://www.algolia.com/docsearch"
+      href={`https://www.algolia.com/ref/docsearch/?utm_source=docsearch&utm_medium=link&utm_term=footer&utm_campaign=${window.location.hostname}`}
       target="_blank"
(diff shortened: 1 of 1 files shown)


packages/docsearch-react/src/AlgoliaLogo.tsx
```diff
@@ -10,9 +10,10 @@ type AlgoliaLogoProps = {
 
 export function AlgoliaLogo({ translations = {} }: AlgoliaLogoProps) {
   const { searchByText = 'Search by' } = translations;
+
   return (
     <a
-      href="https://www.algolia.com/docsearch"
+      href={`https://www.algolia.com/ref/docsearch/?utm_source=docsearch&utm_medium=link&utm_term=footer&utm_campaign=${window.location.hostname}`}
```
bodinsamuel, line 16: As mentionned in DM:
source should be the hostname
medium is referral 
campaign is docsearch
term can be replaced by content (but it's almost equivalent)

bodinsamuel: utm are basically free form query params, but there are more logical way of using them. 
it will affect reporting, but since it was broken we got a clean slate anyway.

shortcuts: Ok make sense, I did not know about that, fixed the params in https://github.com/algolia/docsearch/pull/1259/commits/20f111f633fd8d13e8b8311311d37d8849b19ae8

bodinsamuel (approved):

Issue #1272, opened 2022-01-16
Issue with peer dependencies of `@docsearch/react`


brc-dd:
Ref.: https://github.com/vuejs/vitepress/issues/472#issuecomment-1013835336

---

`@docsearch/js` is using `@docsearch/react` with `preact`, but `@docsearch/react` wants `react`, `react-dom`, `@types/react` as its peer dependencies, which is throwing warnings. `peerDependenciesMeta` can be used to mark react (peer)dependencies in `@docsearch/react` as optional. (It is supported by latest versions of npm, yarn, pnpm.)

brc-dd:
Also, I am not sure why `@algolia/autocomplete-preset-algolia` is listed as dependency of `@docsearch/react`.

`@algolia/autocomplete-preset-algolia` wants `@algolia/client-search` as a peer dependency, but `@docsearch/react` doesn't provide it, this is also giving a warning.

sleeprite:
I also encountered this problem

brc-dd:
> Also, I am not sure why `@algolia/autocomplete-preset-algolia` is listed as dependency of `@docsearch/react`.
> 
> `@algolia/autocomplete-preset-algolia` wants `@algolia/client-search` as a peer dependency, but `@docsearch/react` doesn't provide it, this is also giving a warning.

Ok, so this was fixed by eb372f2743e9f74e2546cde59e676c32b903bece. The peer dependency warnings still exist. #1273 should fix this.

PR #1273, opened 2022-01-16
fix(deps): make react an optional peer dependency in `@docsearch/react`

brc-dd:
fixes (partially) #1272


brc-dd:
Hi @shortcuts! Sorry for pinging you directly. But can this be looked into?

benmccann:
@brc-dd it looks like this PR needs to be rebased

@shortcuts would you be able to review this PR? It's a major stumbling block to using Algolia with non-React projects like Vue, Svelte, etc. Are there any concerns you have or things we could help test?

packages/docsearch-react/package.json
```diff
@@ -42,5 +42,16 @@
...
+  "peerDependenciesMeta": {
+    "@types/react": {
+      "optional": true
+    },
+    "react": {
+      "optional": true
+    },
+    "react-dom": {
+      "optional": true
+    }
```
shortcuts, line 55: Have you tried applying this change at the `@docsearch/js` package level? I believe it wouldn't work but I'd rather not impact this one with unrelated changes.

Ultimately, I believe it's worth the workaround to prevent warnings on non-React packages

shortcuts (commented):
Thanks for the PR! I'm fine with the workaround if we can't apply it at the JS package level

packages/docsearch-react/package.json
brc-dd: Yeah it won't work. `peerDependenciesMeta` can define things "related to the dependencies listed in the `peerDependencies` field". So this workaround needs to be there in the package where we have issue causing peer deps.

brc-dd: Also as you wrote at #1416, we need to make `@algolia/client-search` optional here: https://github.com/algolia/autocomplete/blob/next/packages/autocomplete-preset-algolia/package.json to fix all deps warnings (and work without error on Yarn v2/v3 and PNPM).

shortcuts: > So this workaround needs to be there in the package where we have issue causing peer deps.

I think the workaround is worth it

> Also as you wrote at #1416, we need to make `@algolia/client-search` optional here: https://github.com/algolia/autocomplete/blob/next/packages/autocomplete-preset-algolia/package.json to fix all deps warnings (and work without error on Yarn v2/v3 and PNPM).

Indeed, but actually the issue is located at the Algolia JavaScript API client level (https://github.com/algolia/algoliasearch-client-javascript/issues/1080) not re-exporting types, so Autocomplete shouldn't fix it. We've fixed it in the next major of the API client along with many other changes.

shortcuts (approved):
Thanks for the PR and investigation!

benmccann:
Thanks for the merge! Any chance we can get a new release?

shortcuts:
> Thanks for the merge! Any chance we can get a new release?

Hey, it's released in 3.2.1!

SimenB:
@shortcuts are you sure? 😅 

https://www.npmjs.com/package/@docsearch/react
image

shortcuts:
> @shortcuts are you sure? 😅 
> 
> 
> 
> https://www.npmjs.com/package/@docsearch/react
> 
> image
> 
> 

Oops, now it's good!

PR #1299, opened 2022-02-09
feat(v3): require appId

shortcuts:
## Summary

Follow up of [our new infrastructure](https://docsearch.algolia.com/docs/migrating-from-legacy), users now have their own `appId`, which will be required in v3 stable.

This should be the last breaking change with v3 stable.


bodinsamuel (approved):
lezgongue

PR #1328, opened 2022-03-17
docs: document `aggregateContent`, make API ref clearer

shortcuts:
## Summary

This PR documents the new `aggregateContent` option available on the crawler.

This option allows users to reduce the number of records outputted by the crawler by aggregating all the records of type `content` in a single record.

### DIFF
--- a/packages/website/docs/record-extractor.md
+++ b/packages/website/docs/record-extractor.md
@@ -41,11 +41,9 @@ recordExtractor: ({ helpers }) => {
 },
 ```
 
-## Complex extractors
+### Manipulate the DOM with Cheerio
 
-### Using the Cheerio instance (`$`)
-
-We provide a [`Cheerio instance ($)`][14] for you to retrieve or remove content from the DOM:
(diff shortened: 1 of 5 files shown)


packages/website/docs/record-extractor.md
```diff
@@ -244,7 +268,24 @@ type CustomVariable =
     };
 ```
 
-Contains values that can be used as [`facetFilters`][15]
+Custom variables are used to [`filter your search`](/docs/DocSearch-v3#filtering-your-search), you can define them in the [`recordProps`](#indexing-content-for-faceting)
+
+## `helpers.docsearch` API Reference
```
bodinsamuel, line 273: Can be merged as-is but belongs into official documentation not here
https://www.algolia.com/doc/tools/crawler/apis/configuration/actions/#parameter-param-recordextractor

bodinsamuel (approved):

packages/website/docs/record-extractor.md
shortcuts: I did not found any mention of it on the Algolia doc so I've only added here for now. To avoid duplicate, a doc should point to the other, feel free to decide!

PR #1335, opened 2022-03-24
fix(button): use kbd instead of span

snxx-lppxx:
Replaced span tag to [kbd](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd).

### DIFF
--- a/packages/docsearch-react/src/DocSearchButton.tsx
+++ b/packages/docsearch-react/src/DocSearchButton.tsx
@@ -50,10 +50,10 @@ export const DocSearchButton = React.forwardRef<
       <span className="DocSearch-Button-Keys">
         {key !== null && (
           <>
-            <span className="DocSearch-Button-Key">
+            <kbd className="DocSearch-Button-Key">
               {key === ACTION_KEY_DEFAULT ? <ControlKeyIcon /> : key}
-            </span>
-            <span className="DocSearch-Button-Key">K</span>
+            </kbd>
+            <kbd className="DocSearch-Button-Key">K</kbd>
           </>
         )}
       </span>
--- a/packages/docsearch-react/src/Footer.tsx
+++ b/packages/docsearch-react/src/Footer.tsx
@@ -57,32 +57,32 @@ export function Footer({ translations = {} }: FooterProps) {
       </div>
       <ul className="DocSearch-Commands">
         <li>
-          <span className="DocSearch-Commands-Key">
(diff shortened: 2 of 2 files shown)


packages/docsearch-react/src/DocSearchButton.tsx
```diff
@@ -50,10 +50,10 @@ export const DocSearchButton = React.forwardRef<
...
-            <span className="DocSearch-Button-Key">
+            <kbd className="DocSearch-Button-Key">
               {key === ACTION_KEY_DEFAULT ? <ControlKeyIcon /> : key}
-            </span>
-            <span className="DocSearch-Button-Key">K</span>
+            </kbd>
+            <kbd className="DocSearch-Button-Key">K</kbd>
           </>
         )}
       </span>
```
Kosai106, line 59: +1 for using semantic HTML. I'd also assume the check for `key` should happen one level higher as there's no point in rendering an empty span I'd imagine.


```tsx
      {key !== null && (
        <span className="DocSearch-Button-Keys">
          <kbd className="DocSearch-Button-Key">
            {key === ACTION_KEY_DEFAULT ? 'Search' : key}
          </kbd>
          <kbd className="DocSearch-Button-Key">K</kbd>
        </span>
      )}
```

shortcuts: It was the case before, but we now use the first span to prevent UI flashes: https://github.com/algolia/docsearch/pull/1033

shortcuts (approved):
Thanks, good idea! I've also added it to the footer keys :)

tnir:
@shortcuts Thank you for the update, but from the following changelog, I could not find which package is updated. That could be just why I am not sure how monorepo handles everything (and this kind of problems) at all 😉 

https://github.com/algolia/docsearch/blob/HEAD/CHANGELOG.md#310-2022-05-18

> `button: use kbd instead of span (#1335) (7459a42)`

shortcuts:
> @shortcuts Thank you for the update, but from the following changelog, I could not find which package is updated. That could be just why I am not sure how monorepo handles everything (and this kind of problems) at all 😉

I believe the confusion on your side is because how the DocSearch library is built. We only implement things in the React package, because the JS one is an alias from React to Preact and does not have any logic in it.

So it's either a change on React, which is automatically reflected on JS, or a CSS change.

tnir:
Okay, I got the package structure now. Thank you for clarification.

PR #1384, opened 2022-05-03
fix(searchbutton): fix react hydration error

### LINKED ISSUE #1242: Server-rendering hydration warning in Next.js
## Description

When `@docsearch/react` is used in Next.js (server-side rendering), React prints the warning:

```text
Warning: Expected server HTML to contain a matching <span> in <span>.
    at span
    at span
    at button
    at eval (webpack-internal:///./node_modules/@docsearch/react/dist/esm/DocSearchButton.js:28:32)
    at DocSearch (webpack-internal:///./node_modules/@docsearch/react/dist/esm/DocSearch.js:34:70)
    at div
    at Home
    at MyApp (webpack-internal:///./pages/_app.js:39:27)
```

## Steps to reproduce

Clone my repo and follow the steps in the README: https://github.com/srmagura/docsearch-hydration-repro

## Expected behavior

There should be no hydration warning.

## Environment

- OS: Windows 11
- Browser: Chrome 96
- DocSearch version: 3.0.0-alpha.42

nissy-dev:
fix #1242 

After upgrading React v18, the error of react hydration mismatch happens not only in dev but also in prod.
I fixed this error by using useEffect, following the Next.js documentation, https://nextjs.org/docs/messages/react-hydration-error.

### DIFF
--- a/packages/docsearch-react/src/DocSearchButton.tsx
+++ b/packages/docsearch-react/src/DocSearchButton.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { ControlKeyIcon } from './icons/ControlKeyIcon';
 import { SearchIcon } from './icons/SearchIcon';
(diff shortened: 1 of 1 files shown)


shortcuts (commented):
Thanks for the PR!

IIRC we switched to `useMemo` to prevent a UI flash, but it was not the solution and did not reverted the change. cc @francoischalifour 

Can you confirm this fixes the production only? Locally I was still getting the error for dev builds

shortcuts (approved):
Thanks!

nissy-dev:
@shortcuts 
> Can you confirm this fixes the production only? Locally I was still getting the error for dev builds

I confirmed both production build and dev build and works well. I got the error at the first too, but realized it was due to the build cache. After removing the `.next` directory and running `next dev` again, the error disappeared and the behavior was as expected.

Issue #985, opened 2020-11-25
[next] Refine typescript support


meteorlxy:
1. Please disable `allowSyntheticDefaultImports`

https://github.com/algolia/docsearch/blob/f30ca386cd52327b4b8bb374e796b293b7cd0133/tsconfig.json#L3

This options is poisonous. When users install `@docsearch/react` in their own project, they have to enbale `allowSyntheticDefaultImports` or enable `skipLib` to make things work. Thus, for library project, it's better not to enbale this kind of compilerOptions.

```sh
node_modules/@docsearch/react/dist/esm/useDocSearchKeyboardEvents.d.ts:1:8 - error TS1259: Module '"node_modules/@types/react/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag

1 import React from 'react';
         ~~~~~

  node_modules/@types/react/index.d.ts:65:1
    65 export = React;
       ~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag.
```

2. Put `@types/react` in dependencies of `@docsearch/react`.

As the error shows, when using `@docsearch/react` with typescript, the types definition of react is also required.

The generated `@docsearch/react/dist/esm/DocSearch.d.ts` has `import React from 'react';` in it.

meteorlxy:
@francoischalifour @shortcuts 

Sorry for bothering you guys. Do you have any plans on this?

shortcuts:
Hi @meteorlxy,

Thanks for opening this issue.

1. DocSearch is based on other Algolia librairies, such as Autocomplete, which also requires the `allowSyntheticDefaultImports` option. To keep our librairies aligned, we should not remove it for now.
2. There's also downsides to move it to the dependencies, such as different React versions in the `node_modules` if the version range doesn't match. One solution could be to define it as optional with the `peerDependenciesMeta` tag but _I think_ it wouldn't warn when `@types/react` isn't installed. As it's very inconsistent, leaving it as-is might be the best option. (wdyt @Haroenv ?)
3. Could you please send us a reproduction of this use case?

meteorlxy:
3. A reproduction is ok, but I think you could figure out the reason once you check the dist type file of `@docsearch/js` 😅 

https://github.com/algolia/docsearch/blob/2c0962ebb399885bafbcf7fd20f13b71675494e7/packages/docsearch-js/package.json#L17

[...]

PR #1396, opened 2022-05-26
fix(js): use `outDir` to avoid ambient type definition creation

IanVS:
Fixes one of the issues reported in https://github.com/algolia/docsearch/issues/985.

This generates the same kind of definition files as `@docsearch/react`, not ambient (global) definitions that were being produced previously, which were incorrect (see reproduction: https://stackblitz.com/edit/typescript-decjve?file=index.ts).

I tested this by copying the new `dist` folder into the `node_modules` of my own project, and the type error disappeared.


IanVS:
Here's a modified CSB based on the bot's above, that shows the issue is resolved: https://codesandbox.io/s/vanilla-forked-s6o0zl?file=/src/index.ts

shortcuts (approved):
Thanks! ☺️

PR #1442, opened 2022-07-19
feat(css): make breakpoint customizable

### LINKED ISSUE #1439: [docsearch-css] Breakpoint is to be customizable
## Describe the problem

Breakpoints are now fixed and not customizable to align with other library, framework, or design systems.

https://github.com/algolia/docsearch/blob/83abe4c120bb7f5ae5bbd6e0faf229873546d768/packages/docsearch-css/src/button.css#L66
https://github.com/algolia/docsearch/blob/83abe4c120bb7f5ae5bbd6e0faf229873546d768/packages/docsearch-css/src/modal.css#L559

## Describe the solution

Use CSS custom properties to set all breakpoint(s) in docsearch-css.

## Alternatives you've considered

As per Apples's products i.e. iPad, we can directly change that with 768px without a custom property.

tnir:
Closes #1439

Signed-off-by: Takuya Noguchi [takninnovationresearch@gmail.com](https://github.com/sponsors/tnir)

### DIFF
--- a/packages/docsearch-css/src/_variables.css
+++ b/packages/docsearch-css/src/_variables.css
@@ -44,6 +44,9 @@
   --docsearch-footer-background: #fff;
   --docsearch-footer-shadow: 0 -1px 0 0 rgb(224, 227, 232),
     0 -3px 6px 0 rgba(69, 98, 155, 0.12);
+
+  /* breakpoints */
+  --docsearch-breakpoint: 750px;
 }
 
 /* Darkmode */
--- a/packages/docsearch-css/src/button.css
+++ b/packages/docsearch-css/src/button.css
@@ -63,7 +63,7 @@
   width: 20px;
 }
 
-@media (max-width: 750px) {
+@media (max-width: var(--docsearch-breakpoint)) {
   .DocSearch-Button-Keys,
   .DocSearch-Button-Placeholder {
     display: none;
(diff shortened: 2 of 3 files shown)


shortcuts:
Hey, thanks for the PR!

Could you please update [this sandbox](https://codesandbox.io/s/vanilla-forked-8vqpq9?file=/src/index.css) to see if it properly work? IIRC media queries are not bound to the DOM so can't read `var`s

shortcuts:
> Should I use https://github.com/csstools/postcss-plugins/tree/main/plugins/postcss-env-function instead?

I don't see other alternatives to do it, but since media queries are applied on classes, I believe overriding it from your source code should be enough, no?

You can set it to do nothing at `750px` and apply your logic for any `px` you'd like to trigger it

tnir:
Thanks. I just wanted to do it easy to just change the breakpoint:

```html
<style>
:root {
  --docsearch-breakpoint: 768px;
}
</style>
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3.1.1/dist/style.min.css">
```
No bundling, delivery of the library (`@docsearch/css`) is required.

shortcuts:
> Thanks. I just wanted to do it easy to just change the breakpoint:

I believe overriding media queries are working, you need to first import our stylesheet, then yours with your own logic in it

tnir:
Thanks, your solution is perfect! 👑

PR #1449, opened 2022-07-21
fix(form): validate URL correctly

bodinsamuel:


packages/website/src/components/ApplyForm.js
```diff
@@ -165,11 +165,11 @@ function ApplyForm() {
           <Input
             required={true}
             id="repoURL"
-            type="repoURL"
+            type="url"
```
shortcuts, line 168: Ah yes make sense, I realized there was some weirdly formatted ones on the hub 🤔

shortcuts (approved):

packages/website/src/components/ApplyForm.js
bodinsamuel: yup a lot, I wasn't sure it was intentional but a few people were taking the PIPE very seriously. Should remove some error 👍🏻

shortcuts: good catch!

PR #1535, opened 2022-09-01
chore: add CODEOWNERS

shortcuts:
## summary

Add CODEOWNERS file

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,5 @@
 # This file list code owners for the repository, or part of it.
 # See https://help.github.com/articles/about-code-owners/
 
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# This file list code owners for the repository, or part of it.
+# See https://help.github.com/articles/about-code-owners/
+
+* @shortcuts @francoischalifour
+* @algolia/events-records-connection
```
shortcuts, line 5:
```suggestion
```

Actually I don't think we should tag the team

francoischalifour (approved):

PR #1547, opened 2022-09-05
docs: add `maxDepth` to templates

shortcuts:
## Summary

https://algolia.atlassian.net/browse/CRAW-2430

Adds the `maxDepth` parameter to the recommended templates.

### DIFF
--- a/packages/website/docs/templates.mdx
+++ b/packages/website/docs/templates.mdx
@@ -4,7 +4,9 @@ title: Config templates
 
 import useBaseUrl from '@docusaurus/useBaseUrl';
 
-To help you create the best search experience for your users, we provide config templates for multiple websites generators. If you'd like to add a new template to our list, or believe we should update an existing one, please [send us an email][1] or [open a pull request][2]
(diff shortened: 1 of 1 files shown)


millotp (approved):
Are we sure this will work for all config ? Otherwise good idea

shortcuts:
> Are we sure this will work for all config ? Otherwise good idea

10 is already pretty big for documentations, would you see it higher? The good point is also that it provides a clear `url_too_deep` error which can be easily identifiable for a user

millotp:
For documentation site it makes sense, maybe we can add a link to the [doc](https://www.algolia.com/doc/tools/crawler/apis/configuration/max-depth/) or the error when it happens, so they know which property to change

shortcuts:
> For documentation site it makes sense, maybe we can add a link to the [doc](https://www.algolia.com/doc/tools/crawler/apis/configuration/max-depth/) or the error when it happens, so they know which property to change

On the DocSearch side? It make sense! Would you be able to provide a PR with what you have in mind and put it on this page? https://docsearch.algolia.com/docs/templates

PR #1483, opened 2022-08-05
Fix: use navigator.userAgentData.platform if available

bershanskiy:
Avoid calling `navigator.platform` on Chromium/Chrome where `navigator.userAgentData.platform` is available. This removes the annoying warning in Chromium "Issues" console which reads:
>Audit usage of navigator.userAgent, navigator.appVersion, and navigator.platform

This PR does not make any real changes to functionality.

PR #1548, opened 2022-09-09
fix(navigator): check for experimental `userAgentData` first

shortcuts:
## Summary

https://algolia.atlassian.net/browse/CRAW-2433

Proposed in https://github.com/algolia/docsearch/pull/1483 by @bershanskiy, moved here to add tests on top

This addition prevents the console warning when trying to detect the platform. We still fallback on the default navigator implementation if not found.


.eslintrc.js
```diff
@@ -68,6 +68,7 @@ module.exports = {
       },
       rules: {
         'jest/expect-expect': 0,
+        'jest/valid-expect': 0,
```
francoischalifour, line 71: Why is this change required globally? We can inline the rule disable comment.

cypress/integration/search/actions.spec.ts
```diff
@@ -138,3 +138,33 @@ describe('Recent and Favorites', () => {
...
   });
 });
+
+describe('Navigator', () => {
+  it(
+    'Use correct navigator to detect Apple device on chrome',
+    { browser: 'chrome' },
+    () => {
+      expect((navigator as any).userAgentData).not.be.undefined;
+      expect(navigator.platform).not.to.be.undefined;
```
francoischalifour, line 148: This is testing the Chrome platform, not DocSearch. Why would we test this?

packages/docsearch-react/src/__tests__/api.test.tsx
```diff
@@ -66,12 +68,13 @@ describe('api', () => {
           }}
         />
       );
+      // We can silence the potentially undefined here because we assert that they are not
```
francoischalifour, line 71: I don't think this comment is necessary, this is the TypeScript contract, we don't need to explain it in our codebase.

.eslintrc.js
shortcuts: this is only scoped to the cypress tests, the assertion is not a jest one but the cypress one

Issue #1731, opened 2022-12-27
Can't focus on the `clear` button with tab key


chenxsan:
## Steps to reproduce

1. Go to https://docsearch.algolia.com/
2. Click on the search button
3. type any characters into the search box
4. Now you should see the `X` button, i.e., `clear` button.
5. Try `tab` key to move focus into that button - you can't

## Expected behavior

I should be able to `tab` into the `clear` button.

## Environment

I've tested it under macOS with Firefox/Edge/Safari, none of them work as I expected.

shanpriyan:
Seems like the `clear (x)` button is getting focused but not visible to the user. When I pressed the `Enter` key after pressing `Tab` key from search box it cleared the search box.

chenxsan:
> Seems like the `clear (x)` button is getting focused but not visible to the user. When I pressed the `Enter` key after pressing `Tab` key from search box it cleared the search box.

Exactly, the outline of `:focus` state was disabled https://github.com/algolia/docsearch/blob/main/packages/docsearch-css/src/modal.css#L161.

shortcuts:
Hey, thanks both! released in 3.3.1

PR #1732, opened 2022-12-27
fix(modal): enable outline for focus state

chenxsan:
Closes https://github.com/algolia/docsearch/issues/1731

Users of DocSearch can always customize as needed.


shortcuts:
can confirm it works well on the playground, for both JS and React, looks much better!

Screenshot 2022-12-27 at 12 25 19

shortcuts (approved):
Thanks a lot for spotting the issue and fixing it!

Issue #1370, opened 2022-04-23
Accessibility issues with v3 dialog


patrickhlauke:
## Description

Taking the vanilla v3 playground https://codesandbox.io/s/docsearchjs-v3-playground-z9oxj as an example, there's a few accessibility-related oddities and issues that should be corrected. There's likely more, but a few things that caught my eye right away:

### Incorrect role/attributes on dialog 

Once the search dialog appears, the container for the search currently has `role="button" aria-expanded="true" aria-haspopup="listbox"` and is made focusable using `tabindex="0"`.

The container should instead have `role="dialog" aria-modal="true"`. It can't be a "button" because buttons can't then have focusable controls inside them. 

While focus has been made to cycle-remain inside the dialog, it's still possible for a keyboard user to jump back out and into the underlying page (navigating/setting focus through the page behind the semi-transparent overlay) - for instance, using `Alt+d` to set focus to the browser's address bar and then `Tab`bing forward.

![Screenshot showing focus on the 'Search' control behind the semi-transparent overlay of the example dialog](https://user-images.githubusercontent.com/895831/164895121-89c8c048-b6af-42f9-9aa2-cb781e3326db.png)

Likewise, the underlying page is still exposed to assistive technologies (i.e. screen readers). Using `aria-modal="true"` most modern ATs will ignore anything outside of the dialog, but for older browser/AT combinations, you'd probably want to use `aria-hidden="true"` on the underlying page (for this to work properly though, the modal container itself must live outside of the element that you've set to `aria-hidden="true"`, otherwise it will be hidden as well - and `aria-hidden` can't be overridden in a child element by using `aria-hidden="false"` either).

Video using Chrome/NVDA on Windows: after opening the dialog, exit form entry mode (`NVDA + Space`) then use reading keys (cursor keys) to navigate right back out of the dialog and into the underlying page, as it's still exposed to AT.

https://user-images.githubusercontent.com/895831/164896236-ea2780b2-a1d0-4543-ac76-6f90fbbbf0f7.mp4

One way to work around this would be to either use the new `<dialog>` element (which takes care of most of these), or to use the `inert` attribute and relevant polyfill https://github.com/WICG/inert

### Incorrect role/attributes on the search input itself

[...]

PR #1735, opened 2022-12-28
feat: restore focus after closing modal

chenxsan:
Related issue https://github.com/algolia/docsearch/issues/1370

cc @patrickhlauke if you don't mind please help review this one.

Basically the code would store the `document.activeElement` when modal is opened, and call `focus` on that stored `document.activeElement` when modal is closed.

### DIFF
--- a/examples/demo/src/App.js
+++ b/examples/demo/src/App.js
@@ -8,6 +8,7 @@ function App() {
   return (
     <div>
       <h1>DocSearch v3 - React</h1>
+      <button>A button</button>
       <DocSearch
         indexName="docsearch"
         appId="R2IYF7ETH7"
--- a/packages/docsearch-react/src/DocSearch.tsx
+++ b/packages/docsearch-react/src/DocSearch.tsx
@@ -42,11 +42,12 @@ export interface DocSearchProps {
   initialQuery?: string;
   navigator?: AutocompleteOptions<InternalDocSearchHit>['navigator'];
   translations?: DocSearchTranslations;
-  getMissingResultsUrl?: ({ query: string }) => string;
+  getMissingResultsUrl?: ({ query }: { query: string }) => string;
 }
 
 export function DocSearch(props: DocSearchProps) {
   const searchButtonRef = React.useRef<HTMLButtonElement>(null);
+  const activeElementRef = React.useRef<Element>(document.body);
   const [isOpen, setIsOpen] = React.useState(false);
   const [initialQuery, setInitialQuery] = React.useState<string | undefined>(
     props?.initialQuery || undefined
@@ -54,9 +55,15 @@ export function DocSearch(props: DocSearchProps) {
 
   const onOpen = React.useCallback(() => {
     setIsOpen(true);
+    activeElementRef.current = document.activeElement || document.body;
   }, [setIsOpen]);
 
   const onClose = React.useCallback(() => {
+    if (activeElementRef.current) {
+      if (activeElementRef.current instanceof HTMLElement) {
+        activeElementRef.current.focus();
+      }
+    }
     setIsOpen(false);
   }, [setIsOpen]);
 
--- a/examples/js-demo/index.html
+++ b/examples/js-demo/index.html
@@ -11,7 +11,7 @@
   <body>
     <div class="container">
       <h1>DocSearch v3 - Vanilla JavaScript</h1>
-
+      <button>A button</button>
       <div id="docsearch"></div>
     </div>
 


patrickhlauke:
is there an easy way for me to check this live without having to work out how to run it locally?

shortcuts:
Thanks a lot for the PR!

> is there an easy way for me to check this live without having to work out how to run it locally?

For a concrete test you'd have to run de playground locally, it should only require installing deps and run `yarn playground:start`

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -42,21 +42,28 @@ export interface DocSearchProps {
   initialQuery?: string;
   navigator?: AutocompleteOptions<InternalDocSearchHit>['navigator'];
   translations?: DocSearchTranslations;
-  getMissingResultsUrl?: ({ query: string }) => string;
+  getMissingResultsUrl?: ({ query }: { query: string }) => string;
```
shortcuts, line 45: Oops good catch

shortcuts (commented):
Thanks for tackling this, would you mind adding a cypress test for this case?

chenxsan:
> Thanks a lot for the PR!
> 
> > is there an easy way for me to check this live without having to work out how to run it locally?
> 
> For a concrete test you'd have to run de playground locally, it should only require installing deps and run `yarn playground:start`

I can see there's a netlify deploy preview, but it's canceled, any chance to make it work so anyone can test it online instead of running it locally?

chenxsan:
> Thanks for tackling this, would you mind adding a cypress test for this case?

Sure, I'll add one.

shortcuts:
> > Thanks a lot for the PR!
> 
> > 
> 
> > > is there an easy way for me to check this live without having to work out how to run it locally?
> 
> > 
> 
> > For a concrete test you'd have to run de playground locally, it should only require installing deps and run `yarn playground:start`
> 
> 
> 
> I can see there's a netlify deploy preview, but it's canceled, any chance to make it work so anyone can test it online instead of running it locally?

You can remove the ignore build command of the netlify.toml file, but I don't know if we have a button to re-focus on the website.

chenxsan:
Converted into draft until I figure out the cypress test here.

chenxsan:
@shortcuts Seems we can't test this feature against the website as [docusaurus](https://github.com/facebook/docusaurus) customizes its search function based on `DocSearch` per https://github.com/facebook/docusaurus/blob/0985fa0af338068b4906f409a3f874a319f5359c/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx#L148.

chenxsan:
> is there an easy way for me to check this live without having to work out how to run it locally?
> 
> from looking over the code itself, this looks correct though.

FYI, if you're going to run locally, you would need Node 16. 18 won't work as far as I see. Here's how you would run the project locally:

1. clone
2. run `yarn`
3. run `yarn build`
4. run `yarn playground:start`

shortcuts:
> Hence I would suggest running cypress against the example site, either react-example or js-example should be ok. Testing against a customized search function based on DocSearch where we have no control does't make sense to me.

Makes a lot of sense indeed! Changing the `website:test` here https://github.com/algolia/docsearch/blob/main/package.json#L24 with `playground:start` should solve this, no?

shortcuts:
(again thanks a lot for the time spent here!)

examples/demo/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "build": "parcel build index.html",
-    "start": "parcel index.html"
+    "start": "parcel index.html --port 3000 --no-cache"
```
chenxsan, line 9: I found the code just won't update when developing locally even I was running `yarn watch` under the project root. Not sure what's wrong under the hood so I disable the cache to work around it.

examples/demo/src/App.js
```diff
@@ -8,10 +8,22 @@ function App() {
   return (
     <div>
       <h1>DocSearch v3 - React</h1>
+      <button data-testid="btn">A button</button>
       <DocSearch
         indexName="docsearch"
         appId="R2IYF7ETH7"
         apiKey="599cec31baffa4868cae4e79f180729b"
+        transformItems={
+          (items) => { // transform absolute url into relative ones to solve CORS problem in cypress
```
chenxsan, line 17: The api would return urls of different origin from `localhost` which would result in a CORS error in cypress, so I transform the results. It's a simplify version of how docusaurus makes it https://github.com/facebook/docusaurus/blob/0985fa0af338068b4906f409a3f874a319f5359c/packages/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx#L175

package.json
```diff
@@ -21,7 +21,7 @@
     "cy:run:chrome": "yarn run cy:run --browser chrome",
     "cy:run:edge": "yarn run cy:run --browser edge",
     "cy:run:firefox": "yarn run cy:run --browser firefox",
-    "cy:run": "start-server-and-test 'yarn website:test' http://localhost:3000 'cypress run --spec 'cypress/integration/**/*' --headless'",
+    "cy:run": "start-server-and-test 'yarn playground:start' http://localhost:3000 'cypress run --spec 'cypress/integration/**/*' --headless'",
```
chenxsan, line 24: I'm testing against the React version of demo, not the js version due to some problems I ran into when using the latter version like `yarn watch` won't output umd version. Anyway, could be changed later once those problems get fixed.

cypress/integration/search/actions.spec.ts
```diff
@@ -3,6 +3,9 @@
 describe('Start', () => {
   beforeEach(() => {
     cy.visit(Cypress.config().baseUrl!);
+    // seems we could run into race condition
+    // if we don't wait for .DocSearch-Button-Key to render
+    cy.get('.DocSearch-Button-Key');
```
chenxsan, line 8: I found the tests became flaky without this.

shortcuts: Ah great catch!

chenxsan:
> (again thanks a lot for the time spent here!)

It's my pleasure to improve a project I use every day :)

I think it's ready for review now, I've added some comments to explain why I made such change.

examples/demo/src/App.js
shortcuts: Can we maybe wrap it in an env check? So it's only replaced for the tests

Or you can copy the whole demo folder and name it like `cypress`

examples/js-demo/app.js
```diff
@@ -8,4 +8,13 @@ docsearch({
...
   apiKey: '599cec31baffa4868cae4e79f180729b',
+  transformItems(items) { // transform absolute url into relative ones to solve CORS problem in cypress
+    return items.map((item) => {
+      const { origin } = new URL(item.url);
+      return {
+        ...item,
+        url: item.url.replace(new RegExp(`^${origin}`), ''),
+      };
+    });
+  },
```
shortcuts, line 19: we only use the react demo so this one can stay right?

package.json
shortcuts: No problem it's more important to test the react version since it's where all the logic belong

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -42,24 +42,36 @@ export interface DocSearchProps {
...
   }, [setIsOpen]);
 
   const onClose = React.useCallback(() => {
     setIsOpen(false);
   }, [setIsOpen]);
 
+  useEffect(() => {
+    if (!isOpen) {
+      if (activeElementRef.current) {
+        if (activeElementRef.current instanceof HTMLElement) {
```
shortcuts, line 68: not a big deal but we can wrap it in a single if statement

shortcuts (commented):
Looks great! Thanks a lot

examples/demo/src/App.js
chenxsan: Good idea, I'd go with another folder so we can avoid messing up with other two demo.

packages/docsearch-react/src/DocSearch.tsx
chenxsan: done.

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -4,7 +4,7 @@ import type {
 } from '@algolia/autocomplete-core';
 import type { SearchOptions } from '@algolia/client-search';
 import type { SearchClient } from 'algoliasearch/lite';
-import React from 'react';
+import React, { useEffect } from 'react';
```
shortcuts, line 7: can you use `React.useEffect` instead? just so it's consistent with the rest

shortcuts (commented):
last comment!

chenxsan:
Ooops, CI failed due to https://support.circleci.com/hc/en-us/articles/115014359648-Exit-Code-137-Out-of-Memory

shortcuts:
> Ooops, CI failed due to https://support.circleci.com/hc/en-us/articles/115014359648-Exit-Code-137-Out-of-Memory

oops D: you can update https://github.com/algolia/docsearch/pull/1735/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L18 to prevent building the playground, we don't need it on the CI anyway

I think adding `--ignore @docsearch/(cypress|react-example|js-example)` should do it

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -42,24 +42,36 @@ export interface DocSearchProps {
   initialQuery?: string;
   navigator?: AutocompleteOptions<InternalDocSearchHit>['navigator'];
   translations?: DocSearchTranslations;
-  getMissingResultsUrl?: ({ query: string }) => string;
+  getMissingResultsUrl?: ({ query }: { query: string }) => string;
 }
 
 export function DocSearch(props: DocSearchProps) {
   const searchButtonRef = React.useRef<HTMLButtonElement>(null);
+  const activeElementRef = React.useRef<Element>(document.body);
```
francoischalifour, line 50: Is this working in server environments where `document` isn't defined?

francoischalifour (commented):
Thank you @chenxsan!

packages/docsearch-react/src/DocSearch.tsx
chenxsan: Interesting, I've never thought of rendering `DocSearch` in server side.

I guess it won't work.

Any suggestion?

chenxsan: How about this:



I've reran the tests locally, seems fine.

francoischalifour: Isn't the effect covered by the `onOpen()` callback?

I think we better have `activeElementRef` accept `null` as initial value, and update its value in the `onOpen()` and `onClose()` callbacks. This way, the effect can also be removed, and there's no `document` in the initial render path.

```diff
 export function DocSearch(props: DocSearchProps) {
   const searchButtonRef = React.useRef<HTMLButtonElement>(null);
+  const activeElementRef = React.useRef<Element | null>(null);
   const [isOpen, setIsOpen] = React.useState(false);
   const [initialQuery, setInitialQuery] = React.useState<string | undefined>(
     props?.initialQuery || undefined
   );

   const onOpen = React.useCallback(() => {
     setIsOpen(true);
+    activeElementRef.current = document.activeElement || document.body;
   }, [setIsOpen]);

   const onClose = React.useCallback(() => {
     setIsOpen(false);
+
+    if (activeElementRef.current) {
+      activeElementRef.current.focus();
+      activeElementRef.current = null;
+    }
   }, [setIsOpen]);

-  React.useEffect(() => {
-    if (
-      !isOpen &&
-      activeElementRef.current &&
-      activeElementRef.current instanceof HTMLElement
-    ) {
-      activeElementRef.current.focus();
-    }
-  }, [isOpen]);

   // ...
 }
```

chenxsan: Unfortunately this won't work,

```
   const onClose = React.useCallback(() => {
     setIsOpen(false);
+
+    if (activeElementRef.current) {
+      activeElementRef.current.focus();
+      activeElementRef.current = null;
+    }
   }, [setIsOpen]);
```

I believe I tried this in the beginning, and it would cause two tests to fail:



I.e., clicking the button to open dialog, then invoking shortcut to close the dialog would fail.

PR #1741, opened 2022-12-31
chore: add packageManager field

Mister-Hope:


francoischalifour (commented):
The `packageManager` field [seems to be experimental](https://nodejs.org/api/packages.html#packagemanager) and we don't use it in our other repos, so it might be a bit early.

shaneafsar:
> Also, shall I ask something related to docsearch? I am deploying a monorepo project to a series of sub domains, as mentioned in the docs https://docsearch.algolia.com/docs/DocSearch-program#can-i-have-multiple-projects-under-the-same-algolia-application, I shall apply them one by one, but there are dozens of sites which I am using, and our team ([VuePress](https://v2.vuepress.vuejs.org)) are building a marketplace center at https://vuejs.press and we hope we can let users directly add their docs by applying our subdomain, so can we apply a whole domain for this situation? I try to open a support ticket at algolia but I am not getting any response.
> 
> Here are my open source project sites: https://github.com/vuepress/marketplace/blob/main/config/vuejs.press.zone
> 
> I already tried to contact support but I am not getting any response in days.

Hey @Mister-Hope , do you still need help with adding subdomain support?

Mister-Hope:
Yes!

millotp:
Upgraded to yarn berry in this PR https://github.com/algolia/docsearch/pull/2335, with the `packageManager` field only in the root `package.json`

PR #1742, opened 2022-12-31
feat(js): add destroy method, close #1363

### LINKED ISSUE #1363: Key listenter setting and issues in SPA
Hi, this is Mr.Hope from VuePress Team. 

Docsearch is not providing a method to destory itself or any way to customize (or remove) the search key listener in docs. So I wonder if there is a way, otherwise this will trigger error in SPA application.

Visit https://mrhope.site/ and scroll down then click the first arrticle. When you click `Ctrl + K`, you will find 2 docsearch popup is displayed.

In VuePress, we are using SPA have "layout" comopnent and different page may choose different layout. So when you switch to the first article, the whole page is rerendered because they have different layout, so docsearch is reinitialize. But the previous key listener seems still exist, so keyboard actions seems not correct.

Mister-Hope:


Mister-Hope:
I am unfamilar with react, preact so I am not sure if this is correct, maintainers should pay sepecial attention to my code, thanks.

packages/docsearch-js/src/docsearch.tsx
```diff
@@ -18,18 +18,36 @@ interface DocSearchProps extends DocSearchComponentProps {
...
+  const destroy = useCallback(() => {
+    setShouldRender(false);
+  }, []);
+
+  return { shouldRender, destroy };
+}
+
+// eslint-disable-next-line react-hooks/rules-of-hooks
 export function docsearch(props: DocSearchProps) {
+  const { shouldRender, destroy } = useRender();
```
8bittitan, line 32: @Mister-Hope I know this PR has been sitting for awhile now, but it doesn't seem like this is the correct approach to fixing the issue. A Preact hook has to be used within a Preact component but `docsearch` is just a wrapper function we expose for being able to render.

8bittitan (changes requested):
@Mister-Hope I know this PR has been sitting for awhile now, but it doesn't seem like this is the correct approach to fixing the issue. A Preact hook has to be used within a Preact component but docsearch is just a wrapper function we expose for being able to render.

Mister-Hope:
I am not familiar with react, so you can open a new pr with a valid fix.

xav-ie:
I have made steps in above PR to do what is desired. It now returns unmount function in the preact way of doing things. I will try my best in that PR to fix things if it is not correct.

PR #1781, opened 2023-02-06
fix: Correctly type `getMissingResultsUrl` query as `string`

askoufis:
`({ query: string }) => string;` was being interpreted as a prop rename, which doesn't make sense in the context of an interface, so I've corrected it to what I assume was the intended outcome where `query` is of type `string`, rather than being inferred as `any`.

NOTE: Technically this is a breaking change as `query` was being inferred with type `any`, so consumers that were using it as such may need to change their code. On the other hand, I doubt many consumers are using this value as if it were anything _but_ a `string`, so IMO a patch fix is also valid in this case. I'll leave this decision to the maintainers.


francoischalifour (approved):
Thanks! I think that's fine to release as a minor as it was not intended.

shortcuts:
oops thanks! indeed it was also provided in https://github.com/algolia/docsearch/pull/1735 and not intended D:

Issue #1887, opened 2023-05-16
chrome autocomplete in forms triggers 'keydown' but the event.key is undefined


powerfulyang:
## Description

chrome autocomplete in forms triggers 'keydown' but the event.key is undefined

## Steps to reproduce

<img width="846" alt="image" src="https://github.com/algolia/docsearch/assets/21997724/0237f454-f167-46c7-97d0-2da96150a8d3">

PR #1888, opened 2023-05-16
fix: event.key is undefined #1887

powerfulyang:


shortcuts (commented):
is this for RSC or SSR purposes? looking at the types, it can't be undefined, but if this check solves it, it's not a huge deal to me

cc @francoischalifour wdyt?

powerfulyang:
chrome autocomplete in forms triggers 'keydown' but the event.key is undefined

powerfulyang:
Maybe `KeyboardEvent` type is incorrect ~

packages/docsearch-react/src/useDocSearchKeyboardEvents.ts
```diff
@@ -39,7 +39,8 @@ export function useDocSearchKeyboardEvents({
       if (
         (event.keyCode === 27 && isOpen) ||
         // The `Cmd+K` shortcut both opens and closes the modal.
-        (event.key.toLowerCase() === 'k' && (event.metaKey || event.ctrlKey)) ||
+        (event.key?.toLowerCase() === 'k' &&
```
francoischalifour, line 42: @powerfulyang Can you add a comment saying that there's a bug in Chrome where `event.key` can be `undefined`, and link to https://github.com/paperjs/paper.js/issues/1398?

francoischalifour (commented):
Thanks!

packages/docsearch-react/src/useDocSearchKeyboardEvents.ts
powerfulyang: [scrnli_2023_5_16 20-27-03.webm](https://github.com/algolia/docsearch/assets/21997724/c64dc193-d88b-4d2d-ab95-ef1f45103d9c)
I'm sorry, it's my fault. Watching this video may make things clearer.

powerfulyang:
[scrnli_2023_5_16 20-27-03.webm](https://github.com/algolia/docsearch/assets/21997724/89220237-65f1-437c-be83-b8fe5e7b926e)
the auto-filling from chrome which registers as an undefined key

packages/docsearch-react/src/useDocSearchKeyboardEvents.ts
francoischalifour, line 42: I suppose this would explain what can be going wrong?

```suggestion
        // We need to check for `event.key` because it can be `undefined` with
        // Chrome's autofill feature.
        (event.key?.toLowerCase() === 'k' &&
```

francoischalifour: I'm surprised that this happens in DocSearch though, because DocSearch shouldn't let the Chrome autofill behavior trigger. Do you have a reproduction where I can see it happen with DocSearch?

powerfulyang: Not DocSearch itself.
https://github.com/algolia/docsearch/blob/bae117ce6db0e51f7c0958215705dc849b25212c/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#LL67C51-L67C51
`keydown` is listening globally, so my component will cause this issue.

francoischalifour: Ah got it. You can accept my suggestion above and I'm good with that change 👍

powerfulyang:
English is not my native language; please forgive me if I offend you in any way.

francoischalifour (approved):

francoischalifour:
@powerfulyang Nope, nothing to worry about 🙂 Thanks for your contribution!

shortcuts (approved):
Thanks a lot for the contribution!

filipw01:
For anyone looking for the actual chrome issue https://issuetracker.google.com/issues/41425904

PR #1889, opened 2023-05-16
refactor(searchbox): replace deprecated 'keyCode' with 'key'

letavocado:
Also needs to be replaced [here(#L61-L63)](https://github.com/letavocado/docsearch/blob/eeba7b7e3a92dd08c657b9840869f30f7334d628/packages/docsearch-react/src/useDocSearchKeyboardEvents.ts#L61-L63). need help
```js
if (null.test(String.fromCharCode(event.keyCode))) {
  onInput(event);
}
```


millotp:
sorry I didn't see this PR, this was in https://github.com/algolia/docsearch/pull/2326

PR #1891, opened 2023-05-17
feat: implement `maxResultsPerGroup` prop

bojanrajh:
This PR implements the `maxResultsPerGroup` prop, allowing us to customize the number of displayed results in a group (previously hardcoded to 5, changed to 7 on the example below).

```html
<DocSearch ... maxResultsPerGroup="7" />
```

image


shaneafsar (approved):
Thank you for the contribution!

PR #1904, opened 2023-05-30
feat(api): allow insights options to be forwarded to Autocomplete

8bittitan:
This PR updates the version of `autocomplete-core` that we are using to allow for sending events within DocSearch.

Allowing for events to be sent will unlock more features in the long run for DocSearch and can help give insight into how users are using DocSearch on each site.

### DIFF
--- a/packages/docsearch-react/src/DocSearchModal.tsx
+++ b/packages/docsearch-react/src/DocSearchModal.tsx
@@ -130,8 +130,6 @@ export function DocSearchModal({
     [favoriteSearches, recentSearches, disableUserPersonalization]
   );
 
-  const insightsActive = Boolean(insights);
-
   const autocomplete = React.useMemo(
     () =>
       createAutocomplete<
@@ -196,6 +194,8 @@ export function DocSearchModal({
               },
             ];
           }
+
+          const insightsActive = Boolean(insights);
 
           return searchClient
             .search<DocSearchHit>([
@@ -346,7 +346,6 @@ export function DocSearchModal({
       transformItems,
       disableUserPersonalization,
       insights,
-      insightsActive,
       appId,
       apiKey,
     ]
--- a/packages/docsearch-react/src/DocSearchModal.tsx
+++ b/packages/docsearch-react/src/DocSearchModal.tsx
@@ -58,6 +58,7 @@ export function DocSearchModal({
   initialQuery: initialQueryFromProp = '',
   translations = {},
   getMissingResultsUrl,
+  insights = false,
 }: DocSearchModalProps) {
   const {
     footer: footerTranslations,
@@ -147,10 +148,12 @@ export function DocSearchModal({
             searchSuggestions: [],
           },
         },
+        insights,
         navigator,
         onStateChange(props) {
           setState(props.state);
         },
+        // @ts-expect-error
         getSources({ query, state: sourcesState, setContext, setStatus }) {
           if (!query) {
             if (disableUserPersonalization) {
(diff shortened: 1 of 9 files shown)


packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -147,10 +150,12 @@ export function DocSearchModal({
             searchSuggestions: [],
           },
         },
+        insights,
         navigator,
         onStateChange(props) {
           setState(props.state);
         },
+        // @ts-expect-error
```
francoischalifour, line 158: It's dangerous to ignore typing errors for such a foundational function. What error are you expecting here?

8bittitan: @francoischalifour sorry this is still WIP so this will go away soon 😅

8bittitan: This has been updated and removed now :)

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -285,16 +303,23 @@ export function DocSearchModal({
...
+                            const potentialParent = groupedHits.find(
+                              (siblingItem) =>
+                                siblingItem.type === 'lvl1' &&
+                                siblingItem.hierarchy.lvl1 ===
+                                  item.hierarchy.lvl1
+                            ) as InternalDocSearchHit;
+
+                            if (item.type !== 'lvl1' && potentialParent) {
+                              parent = potentialParent;
+                            }
```
francoischalifour, line 317: What's this change doing? How is it related to Insights?

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -129,6 +130,8 @@ export function DocSearchModal({
     [favoriteSearches, recentSearches, disableUserPersonalization]
   );
 
+  const insightsActive = Boolean(insights);
```
francoischalifour, line 133: We could inline `Boolean(insights)` in `getSources` and get rid of this `useEffect` dep, we already have `insights` in the deps.

packages/docsearch-react/src/DocSearchModal.tsx
8bittitan: This was more of a cleanup to help with the typings of things. Overall no logic changes were made, just working around TypeScript

packages/docsearch-react/src/DocSearchModal.tsx
8bittitan: Oh good callout! Moved 😄

shaneafsar (approved):

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -285,16 +303,23 @@ export function DocSearchModal({
...
                         .map((groupedHits) =>
                           groupedHits.map((item) => {
+                            let parent: InternalDocSearchHit | null = null;
+
+                            const potentialParent = groupedHits.find(
+                              (siblingItem) =>
+                                siblingItem.type === 'lvl1' &&
+                                siblingItem.hierarchy.lvl1 ===
+                                  item.hierarchy.lvl1
+                            ) as InternalDocSearchHit;
```
francoischalifour, line 313: This lies a little bit less with the type:

```suggestion
                            ) as InternalDocSearchHit | undefined;
```

francoischalifour (approved):

shortcuts (approved):
looks super great!

(don't forget to update the squash commit message so that the release script can catch it as a `feat` (https://www.conventionalcommits.org/en/v1.0.0/))

Issue #1045, opened 2021-09-15
Click and Conversion Analytics


oliviertassinari:
## Describe the problem

Our team is diving into how we can improve the Algolia search results. Having the access to the clicks and how they change in an A/B test could guide us.

## Describe the solution

We could implement https://www.algolia.com/doc/guides/getting-analytics/search-analytics/advanced-analytics/.

shortcuts:
Hey @oliviertassinari,

Search Insights are not supported by DocSearch yet. We are planning to support [Autocomplete plugins](https://github.com/algolia/autocomplete/tree/next/packages/autocomplete-plugin-algolia-insights) before the v3 stable release, which would make this possible out of the box!

In the meantime, you might find useful informations on the analytics we provide [on the dashboard](https://www.algolia.com/apps/BH4D9OD16A/analytics/overview/material-ui?from=2021-09-08&to=2021-09-14)?

clowder:
> Search Insights are not supported by DocSearch yet. We are planning to support [Autocomplete plugins](https://github.com/algolia/autocomplete/tree/next/packages/autocomplete-plugin-algolia-insights) before the v3 stable release, which would make this possible out of the box!

@shortcuts Is this now possible in v3? There doesn't appear to be a way to pass the insights plugin (created with `createAlgoliaInsightsPlugin`) to the autocomplete instance used by docsearch.

yuhan:
Hi @shortcuts is there an estimate for support autocomplete plugins in `DocSearchModal` or `DocSearch`? Meanwhile, without out-of-box support, what's the best way to send analytics when I want to keep using `DocSearchModal`? Could you please point me to a guide or example of how to implement the analytics events with DocSearchModal? - Neither of the docsearch, autocomplete, or algolia docs is helpful in this case. Thank you!

p.s. I liked how convenient and flexible Algolia has been but it's been a big pain point (soon to be a deal breaker) that I can't manage to get click and conversion analytics.

sarahg:
@shortcuts We're very interested in using click analytics with DocSearch as well. We need to be able to track the "No Clicks" rate in particular as we work on improving result quality.

Is it correct that this is still unsupported? I tried setting it up [like this](https://gitlab.com/gitlab-org/gitlab-docs/-/merge_requests/3362/diffs) but am not seeing anything coming in on the Algolia side.

Seems like https://github.com/algolia/docsearch/issues/1688 is a similar question, but for Docusaurus.

Thank you!

[...]

PR #1915, opened 2023-06-12
(chore): Updates the documentation for insights in JS

8bittitan:
Updates our code snippet of using the `insights` option within the JS package.

Pointed out by: https://github.com/algolia/docsearch/issues/1045#issuecomment-1583044970


shortcuts (approved):
good catch

PR #1916, opened 2023-06-12
feat (api): Adds ability for click insights events to be triggered

8bittitan:
Recently we added insights options for DocSearch in #1904. This PR now also adds in the ability for click analytics to be triggered as well when insights option is active.


packages/docsearch-react/src/utils/buildInsightsClickParams.ts
```diff
@@ -0,0 +1,22 @@
...
+  ClickedObjectIDsAfterSearchParams,
+} from '@algolia/autocomplete-plugin-algolia-insights';
+
+import type { InternalDocSearchHit } from '../types';
+
+export function buildInsightsClickParams(
+  item: InternalDocSearchHit,
+  itemPosition: number
+): InsightsParamsWithItems<ClickedObjectIDsAfterSearchParams> {
+  const insightsItem = item as AlgoliaInsightsHit;
```
levimichael, line 13: Could we run into issues with this assertion?

levimichael (approved):
Cool! 🥳

packages/docsearch-react/src/utils/buildInsightsClickParams.ts
8bittitan: I don't think so, if we have an `insights` plugin active the needed attributes are added to the `item`. All `AlgoliaInsightsHit` type adds is better typings for the `__autocomplete_*` attributes.

packages/docsearch-react/src/Hit.tsx
```diff
@@ -8,5 +8,9 @@ interface HitProps {
 }
 
 export function Hit({ hit, children }: HitProps) {
-  return <a href={hit.url}>{children}</a>;
+  return (
+    <a href={hit.url} onClick={(e) => e.preventDefault()}>
```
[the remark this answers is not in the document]
8bittitan: Crap, that was for debugging 😅 Removed :D

shaneafsar (approved):

packages/docsearch-react/src/utils/buildInsightsClickParams.ts
```diff
@@ -0,0 +1,22 @@
+import type { AlgoliaInsightsHit } from '@algolia/autocomplete-core';
+import type {
+  InsightsParamsWithItems,
+  ClickedObjectIDsAfterSearchParams,
+} from '@algolia/autocomplete-plugin-algolia-insights';
+
+import type { InternalDocSearchHit } from '../types';
+
+export function buildInsightsClickParams(
```
francoischalifour, line 9: It looks like this function content can be inlined in `sendItemClickEvent`?

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -130,6 +131,23 @@ export function DocSearchModal({
...
+      const insightsClickParams = buildInsightsClickParams(
+        item,
+        item.__autocomplete_id
+      );
+
+      state.context.algoliaInsightsPlugin.insights.clickedObjectIDsAfterSearch(
+        insightsClickParams
+      );
+    },
+    [state]
```
francoischalifour, line 148: Watching for plugin changes would be more accurate.

```suggestion
    [state.context.algoliaInsightsPlugin]
```

packages/docsearch-react/src/utils/buildInsightsClickParams.ts
8bittitan: Yeah I agree, originally I had the insights inlined to the click handler so this felt better then. Now that it's broken out this makes sense

packages/docsearch-react/src/DocSearchModal.tsx
8bittitan: Good call

packages/docsearch-react/src/Hit.tsx
francoischalifour, line 12: Looks like the  `preventDefault` call should be removed again.

PR #1946, opened 2023-06-23
docs: add docusaurus-v3 entry

slorber:
We are going to release Docusaurus v3 this summer and need a place to link to the new v3 config.

It is just a copy for v2 for now, but I think it's preferable to do the duplication in case the configs diverge in the future: we don't want to impact v2 users if we do changes only in Docusaurus v3


shortcuts (approved):
thanks! I'll make the update on the Crawler side too

PR #1949, opened 2023-06-23
fix: make AlgoliaLogo take its color from the defined CSS variable

icflorescu:


icflorescu:
Oh, well...

levimichael:
@icflorescu curious why you closed this? Seems like a good and simple fix!

PR #1981, opened 2023-07-14
Clarify RegEx requirements

jwitz:
This section of the doc made it seem like a URL object is required to use RegEx, but it works fine in just a string array.


shaneafsar (approved):
Thank you for adding the clarification!

PR #2097, opened 2023-10-12
docs: add missing backtick to `placeholder` default example

nedjulius:
Adds a missing backtick and fixes the formatting.


shortcuts (approved):
thanks!

PR #2103, opened 2023-10-19
chore(site): Update logo & color

shaneafsar:
- Swap indigo for algolia brand color
- add logo update


dylantientcheu (commented):
LGTM, maybe make the left nav logo smaller on mobile since it overlaps a button.

image

dylantientcheu (approved):

8bittitan:
What do you think about making the logo just (DocSearch logo) x (Algolia logo) without the letter marks? Seems like that would be cleaner but would users know which is which?

PR #2117, opened 2023-11-05
fix: Improve ESM compatibility with Node

lachlancollins:
Hi there, I ran into this issue when migrating the [TanStack](https://github.com/TanStack/tanstack.com) site to Remix v2.2 with the experimental Vite plugin. When attempting to run the server after build, the following error occurs:

```
SyntaxError: Named export 'DocSearch' not found. The requested module '@docsearch/react' is a CommonJS module, which may not support all module.exports as named exports.
```

However, the source code is clearly ESM only. This issue is occurring because without an explicit `"type"` field in `package.json`, Node interprets `.js` extensions as CJS. This error can also be seen reported by publint [here](https://publint.dev/@docsearch/react@3.5.2) and [here](https://publint.dev/@docsearch/js@3.5.2).

This PR adds in the `"type"` field, as well as the `"exports"` field, and passes the publint checks. The `babel.config.js` file had to be renamed with a `.cjs` extension because otherwise it would be interpreted as ESM rather than CJS.


lachlancollins:
Seems this is causing the website build to fail, since @docusaurus/preset-classic uses docsearch and v2 doesn't support ESM... any chance that the website could be bumped to Docusaurus v3?

lachlancollins:
This is far more complicated than I anticipated affecting multiple packages, closing for now.

lachlancollins:
For anyone with a similar problem, we got around it by doing the following:

```
import * as DocSearch from '@docsearch/react';

export function Search(props) {
  return <DocSearch.DocSearch {...props} />
}
```

PR #2193, opened 2024-02-26
fix(docsearch-react): add aria-hidden to SearchIcon and add accessible label to SearchBox input

hbuchel:
There are 2 main changes in this PR:
- **Allows passing `aria-hidden` to the SearchIcon** so it can be hidden to assistive tech users. The unlabelled SVG can be accessed by screenreaders and can also get caught by automated accessibility audits. Since it's decorative, we can safely hide it with `aria-hidden="true"` in `<SearchBox />` and `<SearchButton />`
- **Adds an accessible label, `searchInputLabel`, to the search form's input.** Currently the `<input />` is associated with a `<label>` that does not have an accessible label, only an unlabelled SVG as a child element. This PR adds accessibly hidden text within the `<label>`. I opted to use visually hidden text (using CSS to hide the text but still make it available for assistive tech users) for the label text because it's generally a better practice than reaching for `aria-label`.

Additionally, this PR:
- Updates test to check for the new label text and translation, as well as fixes some TS related linting errors in that file.
- Updates the related docs to show the new translation option for `searchInputLabel`

### DIFF
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -551,6 +551,18 @@ svg.DocSearch-Hit-Select-Icon {
   width: 20px;
 }
 
+/* Hide element accessibly, so that it is still accessible to
+assistive tech users */
+.DocSearch-VisuallyHidden {
+  clip: rect(0 0 0 0);
+  clip-path: inset(50%);
+  height: 1px;
+  overflow: hidden;
+  position: absolute;
+  white-space: nowrap;
+  width: 1px;
+}
+
 /* Responsive */
 @media (max-width: 768px) {
   :root {
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -553,7 +553,7 @@ svg.DocSearch-Hit-Select-Icon {
 
 /* Hide element accessibly, so that it is still accessible to
 assistive tech users */
-.DocSearch-VisuallyHiddenForAccessbility {
+.DocSearch-VisuallyHiddenForAccessibility {
   clip: rect(0 0 0 0);
   clip-path: inset(50%);
   height: 1px;
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -553,7 +553,7 @@ svg.DocSearch-Hit-Select-Icon {
 
 /* Hide element accessibly, so that it is still accessible to
 assistive tech users */
-.DocSearch-VisuallyHidden {
+.DocSearch-VisuallyHiddenForAccessbility {
   clip: rect(0 0 0 0);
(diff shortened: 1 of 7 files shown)


packages/docsearch-css/src/modal.css
```diff
@@ -551,6 +551,18 @@ svg.DocSearch-Hit-Select-Icon {
   width: 20px;
 }
 
+/* Hide element accessibly, so that it is still accessible to
+assistive tech users */
+.DocSearch-VisuallyHidden {
```
shortcuts, line 556: Maybe a more descriptive name like 
```suggestion
.DocSearch-VisuallyHiddenForAccessbility {
```

packages/docsearch-react/src/icons/SearchIcon.tsx
```diff
@@ -1,12 +1,13 @@
 import React from 'react';
 
-export function SearchIcon() {
+export function SearchIcon({ ...rest }) {
```
shortcuts, line 3: not sure which way's best but if it's only for aria-hidden, maybe we can directly define on the svg?

shortcuts (commented):
awesome, thanks a lot for taking the time to improve DocSearch

cc @francoischalifour

packages/docsearch-css/src/modal.css
hbuchel: Thanks, updated in latest to `.DocSearch-VisuallyHiddenForAccessibility`

packages/docsearch-react/src/icons/SearchIcon.tsx
hbuchel: That works for me; looks like it's only used in these two places where it makes sense for both to have it hidden. Updated in latest.

francoischalifour (approved):
Thank you @hbuchel!

Could you please slightly increase the [bundle size thresholds](https://github.com/algolia/docsearch/blob/874e16a5d42e8657e6ab2653e9638cd2282ba408/bundlesize.config.json)?

```diff
 {
   "files": [
     {
       "path": "packages/docsearch-css/dist/style.css",
-       "maxSize": "3 kB"
+       "maxSize": "3.25 kB"
     },
     {
       "path": "packages/docsearch-react/dist/umd/index.js",
-       "maxSize": "22.80 kB"
+       "maxSize": "23 kB"
     },
     {
       "path": "packages/docsearch-js/dist/umd/index.js",
-       "maxSize": "30.70 kB"
+       "maxSize": "31 kB"
     }
   ]
 }
```

hbuchel:
> Thank you @hbuchel!
> 
> Could you please slightly increase the [bundle size thresholds](https://github.com/algolia/docsearch/blob/874e16a5d42e8657e6ab2653e9638cd2282ba408/bundlesize.config.json)?
> 

Updated, thanks!

PR #2256, opened 2024-06-05
Fix: updating logo placement for mobile

randombeeper:
Roger X pointed out an issue with the site on mobile, a few people verified it was an issue, at least with chrome devtools, this change seems to fix the horizontal scrolling issue seen on mobile. Making the change in production to verify.


8bittitan (approved):
LGTM!

PR #2261, opened 2024-06-20
fix(logos): removed logos no longer using algolia

randombeeper:
I've followed up with each of these contacts but for now they're not using Algolia. Let's remove them.


levimichael (approved):
Easy! Thanks James!

PR #2269, opened 2024-07-03
fix(logos): they will not be fixing the search

randombeeper:
Removing their logo then


randombeeper:
@8bittitan can you review this too?

levimichael (approved):
👍

PR #2273, opened 2024-07-10
fix: updating the templates page

randombeeper:
The templates page needs some work. The first thing I saw is that it alluded to being able to create a crawler. As a docsearch user, this is confusing. We automatically create a crawler for the user, additional crawlers get confusing for the user.

Provide additional information about updating templates, as we're talking about templates here.

Verify the current templates, I see the default one isn't quite right.

Given that we use the same template for Docusaurus v2 and v3 combine those.

Add the Starlight template.

### DIFF
--- a/packages/website/docs/templates.mdx
+++ b/packages/website/docs/templates.mdx
@@ -1,25 +1,22 @@
 ---
-title: Config templates
+title: Config Templates
 ---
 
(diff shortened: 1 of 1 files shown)


randombeeper:
Not all the intended changes are in yet, more commits coming.

levimichael (approved):
Good improvements!

PR #2278, opened 2024-07-15
fix #2230 - prevent babel helpers duplication and condense bundle & types

### LINKED ISSUE #2230: ESM output contains many duplicated helpers
## Description

As `@docsearch/react` compiles each of its files in isolation for the ESM output, the module as a whole ends up with a dozen duplicated helpers even in the most simple demonstration.

Edit: This applies for all dependencies of `@docsearch/react` too.

This heavily bloats the bundles of applications which use it, and bloats them needlessly.

## Steps to reproduce

1. Start a Vite (though every bundler would be impacted similarly) React project, pasting the following into `src/main.jsx`:

```js
import React from 'react';
import ReactDOM from 'react-dom/client';
import { DocSearch } from '@docsearch/react';

function App() {
  return (
    <DocSearch
      appId="YOUR_APP_ID"
      indexName="YOUR_INDEX_NAME"
      apiKey="YOUR_SEARCH_API_KEY"
    />
  );
}

ReactDOM.createRoot(document.getElementById('root')).render(
  <React.StrictMode>
    <App />
  </React.StrictMode>,
)
```

3. `npm run build`
4. Open built JS
5. See 10 copies of the iterator helper (`"Invalid attempt to destructure non-iterable instance...."`)

## Expected behavior

That there aren't heavily duplicated helpers throughout the bundles.

This likely means moving to better build tooling, as compiling each module in isolation isn't really viable, nor has it ever been. The asset quality takes a pretty big hit when doing so.

## Environment

- OS: N/A
- Browser: N/A
- DocSearch version: 3.6.0

levimichael:
- Fixes https://github.com/algolia/docsearch/issues/2230, by condensing the ESM bundle into a single file
- Uses`rollup-plugin-dts`  to merge the types into one `index.d.ts` file
- Uses `@rollup/plugin` instead of manually passing the bundle to Babel post-build, and replaces `rollup-plugin-babel` usage


8bittitan:
If we are taking the time to update these, what do you think about moving to something like tsup overall? https://tsup.egoist.dev/

8bittitan:
Should we also do this for the `docsearch-js` implementation as well?

levimichael:
> If we are taking the time to update these, what do you think about moving to something like tsup overall? https://tsup.egoist.dev/

Interesting idea, I'm not sure its worth the effort currently to swap out the bundler. We can revisit this in another PR!

8bittitan (approved):

levimichael:
> Should we also do this for the `docsearch-js` implementation as well?

Updated!

PR #2280, opened 2024-07-19
fix(email): docsearch@algolia.com is being retired

randombeeper:
This is a change by support (this email goes to support and automatically creates a ticket), support is no longer accepting ticket creation via email. Replacing with links to Discord or proper support site. More info can be found at https://algolia.atlassian.net/browse/ZA-268


8bittitan (approved):
LGTM

PR #2309, opened 2024-09-27
fix(hits): default to hierarchy lvl0 if highlights present

shortcuts:
https://algolia.atlassian.net/browse/DI-2986

There seems to be a bug on the indexing side which impacts some DocSearch users. This bug has been reported to me by @rickhanlonii and fixed in https://github.com/reactjs/react.dev/pull/7183 temporarily, but we can globally mitigate this issue which also impacts other DocSearch users such as https://vuejs.org/, https://tailwindcss.com/

We now default to the non highlighted lvl0 in case the highlightResult value is not present

You can try this fix by running `yarn playground:start` on `main` and this branch:
- On main (or just https://vuejs.org/), no section appears (cf screenshot 1
- On this branch (cf screenshot 2)

Screenshot 2024-09-27 at 10 31 16

Screenshot 2024-09-27 at 10 31 53

### DIFF
--- a/examples/demo/src/App.js
+++ b/examples/demo/src/App.js
@@ -9,9 +9,9 @@ function App() {
     <div>
       <h1>DocSearch v3 - React</h1>
       <DocSearch
-        indexName="docsearch"
-        appId="R2IYF7ETH7"
-        apiKey="599cec31baffa4868cae4e79f180729b"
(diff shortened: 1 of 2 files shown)


morganleroi (approved):
Let's go !

PR #2326, opened 2024-10-29
fix: upgrade algoliasearch to v5

millotp:
We can use the new generated client to stay up to date and upgrade packages that depends on docsearch.


bundlesize.config.json
```diff
@@ -6,11 +6,11 @@
     },
     {
       "path": "packages/docsearch-react/dist/umd/index.js",
-      "maxSize": "23 kB"
+      "maxSize": "27 kB"
```
shortcuts, line 9: too bad the bundler doesn't resolve to the minified version :(

shortcuts (approved):
looks good!! gg on your first commit :D

cc @francoischalifour in case you want to double check

PR #2329, opened 2024-10-31
fix: use vite instead of parcel in the examples

millotp:
Parcel would delete some random file like `identity` because it though it was not used, `vite` is smarter.
Also remove unused imports


packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -62,7 +62,8 @@ export function DocSearch(props: DocSearchProps) {
 
   const onClose = React.useCallback(() => {
     setIsOpen(false);
-  }, [setIsOpen]);
+    setInitialQuery(props?.initialQuery);
```
shortcuts, line 65: was this to fix the issue of the character leftover?

shortcuts (approved):
looks incredible, gg

PR #2331, opened 2024-10-31
chore(website): use the local package

millotp:
Now that the new version is published, we can add the website to yarn workspaces again without conflicting version


shortcuts (approved):
Ggggg

PR #2332, opened 2024-11-04
fix(types): export the types folder

millotp:
Some types are not exported since `3.6.1` because of https://github.com/algolia/docsearch/pull/2278, but they are imported in [docusaurus/theme-search-algolia](https://www.npmjs.com/package/@docusaurus/theme-search-algolia) for example.

To make the migration smoother we can export those types again.


Fluf22 (approved):
✅

PR #2333, opened 2024-11-04
chore(deps): upgrade cypress

millotp:
The old version we were using is too flaky, the new one seem to be much faster.


Fluf22 (approved):
🚀

PR #2337, opened 2024-11-04
feat(a11y): add hotkeys to search input aria-label

bvandercar-vt:
Hotkey buttons are not represented to a screen reader, but should be.


millotp (approved):
Thanks for your contribution !

PR #2339, opened 2024-11-04
fix: use npm for release

millotp:
`yarn publish` doesn't exists anymore

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -11,7 +11,7 @@ export default {
     packagesToPublish: packages,
   },
   publishCommand({ tag }) {
-    return `yarn publish --access public --tag ${tag}`;
+    return `npm publish --tag ${tag}`;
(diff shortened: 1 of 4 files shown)


ship.config.js
```diff
@@ -11,7 +11,7 @@ export default {
     packagesToPublish: packages,
   },
   publishCommand({ tag }) {
-    return `yarn publish --access public --tag ${tag}`;
+    return `npm publish --tag ${tag}`;
```
shortcuts, line 14: Isn't there something to put for the token to use it on the ci?

shortcuts (approved):

ship.config.js
millotp: it should automatically pick up the `NPM_AUTH_TOKEN` that's on the CI

PR #2348, opened 2024-11-12
fix(types): use a minimal type for transformSearchClient

millotp:
The `transformSearchClient` type has change in the last release (https://github.com/algolia/docsearch/pull/2326) and is breaking in some situation.
To help with the migration, we can expose a new type that is the bare minimum needed by docsearch.


packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -16,17 +15,23 @@ export type DocSearchTranslations = Partial<{
   modal: ModalTranslations;
 }>;
 
+export type DocSearchTransformClient = {
```
shortcuts, line 18:
```suggestion
// The interface that describes the minimal implementation required for the algoliasearch client, when using the [`transformSearchClient`](https://docsearch.algolia.com/docs/api/#transformsearchclient) option
export type DocSearchTransformClient = {
```

wdyt?

we also need to update the definition here https://docsearch.algolia.com/docs/api/#transformsearchclient

shortcuts (commented):
noice

packages/docsearch-react/src/DocSearch.tsx
millotp: good point

shortcuts (approved):
gg!

PR #2377, opened 2024-11-25
chore: add `--host` option to `playground:start` script in `package.json`

lumirlumir:
Hello, I’ve added the `--host` option to the `playground:start` and `playground-js:start` scripts in `package.json`.

This change enables the playground to work on cross-platform devices like mobile phones or tablets using an IP address.

Additionally, it does not affect the previous functionality. It simply extends the playground’s compatibility to work on various devices.

image

This option is documented in Vite.


shortcuts (approved):
great :) thanks for the contribution!

PR #2398, opened 2024-12-13
docs: update ApplyForm.js

randombeeper:
A user asked a question about this text. I felt like it was misleading since you can apply again if you already have a docsearch application. Users do this when they need to cover another domain and want a separate index.

What we won't accept is an application that covers a domain that has previously been approved. In that scenario, the submitted form has an obvious message that its a duplicate domain.


randombeeper:
@dylantientcheu could you review?

shortcuts (approved):

PR #2413, opened 2024-12-17
fix(ui): compensate for scrollbar

### LINKED ISSUE #2378: UI elements shifting to right when model opens
## Description

When the search bar modal opens, elements beneath the popup shift to the right. This issue occurs because the `DocSearch--active` class is added to the body element, causing the scrollbar to disappear, which in turn shifts the layout.
Code causing this issue

```
React.useEffect(() => {
    document.body.classList.add('DocSearch--active');

    return (): void => {
      document.body.classList.remove('DocSearch--active');

      // IE11 doesn't support `scrollTo` so we check that the method exists
      // first.
      window.scrollTo?.(0, initialScrollY);
    };

    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

instead we can use 
```
  useLayoutEffect(() => {
    // Calculate the scrollbar width to compensate for removed scrollbar
    const scrollBarWidth = window.innerWidth - document.body.clientWidth;
    // Prevent layout shift by adding appropriate margin to the body
    document.body.style.marginRight = `${scrollBarWidth}px`;
     
    return (): void => {
      // Reset body styles
      document.body.style.marginRight = '0px';
    }
  }, []);

```
## Steps to reproduce

For Windows
1. Visit the [`...`](https://react.dev/)
2. Click on search bar
3. Observe how the elements beneath the popup modal shift to the right.

For Mac
1. Change the scrollbar behavior to always visible.
2. Visit the [`...`](https://react.dev/)
3. Click on search bar
4. Observe how the elements beneath the popup modal shift to the right.
image

**Live reproduction:**

https://github.com/user-attachments/assets/6bfa04bc-28f7-49b3-bd8f-83c55c8e8412

## Expected behavior

The elements beneath the popup modal should not shift to the right

## Environment

- OS: Windows / Linux / macOS 
- Browser: All browsers
- DocSearch version: 3.8.0

millotp:
fixes #2378


phanendraguptha:
Hello @shortcuts and @francoischalifour,

Could you please review this and share your thoughts about this PR?

shortcuts (approved):

Issue #2414, opened 2024-12-17
v3.8.1 breaks Sass build


XhmikosR:
## Steps to reproduce

1. Go to `...`
2. Click on `...`
3. Scroll down to `...`

**Live reproduction:**
[
- JavaScript: https://codesandbox.io/s/github/algolia/docsearch/tree/main/examples/js-demo
- React: https://codesandbox.io/s/github/algolia/docsearch/tree/main/examples/demo](https://github.com/twbs/bootstrap/actions/runs/12375551822/job/34540617657?pr=41108)

## Environment

- OS: [e.g. Windows / Linux / macOS / iOS / Android]
- Browser: [e.g. Chrome, Safari]
- DocSearch version: [e.g. 3.0.0]

---

I believe it's https://github.com/algolia/docsearch/commit/f3dc6462569719c36fbd40394f25f64f9d29c160 to blame since it's using the newer `@media` syntax. I'd revert the change to keep compatibility with older but still actively used Sass implementations.

XhmikosR:
Refs https://github.com/twbs/bootstrap/pull/41108

shortcuts:
Hey @XhmikosR thanks for reporting, indeed it's in their doc, seems like a miss from our side, https://github.com/algolia/docsearch/pull/2415 should revert it

shortcuts:
Just released 3.8.2, let me know if it has fixed your issue!

XhmikosR:
Thanks for the quick fix! I'll wait for the next patch version.

BTW you could totally check these things with Stylelint. We do have a
configuration we are using, in case it helps you guys.

PR #2415, opened 2024-12-17
fix(css): older sass supported syntax

shortcuts:
closes https://github.com/algolia/docsearch/issues/2414

> LibSass and older versions of Dart Sass and Ruby Sass don’t support media queries with features written in a [range context](https://www.w3.org/TR/mediaqueries-4/#mq-range-context). They do support other standard media queries.

https://www.w3.org/TR/mediaqueries-4/#mq-range-context
https://sass-lang.com/documentation/at-rules/css/#media


Fluf22 (approved):
👴🏻

PR #2485, opened 2025-02-04
fix(a11y): have unique ids for favorite/recent lists

mxschmitt:
There are two distinct a11y bugs which are related to each-other:

1. The main input `.DocSearch-Input` contains `aria-controls="docsearch-recentSearches-list docsearch-favoriteSearches-list"` which is not compliant because the `recentSearches` or `favoriteSearches` elements do not exist all the time in the DOM. Only if existing favorites or recent searches they should be added - see https://github.com/microsoft/playwright.dev/issues/1628. The fix for this is in https://github.com/algolia/autocomplete/pull/1302. 
2. When there is a _Recent_ and a _Favourite_ results list is shown, they share the same underlying `id="docsearch-list"`. This PR addresses that and adds a test for both issues.

**Investigation**

This seems to render the search list here:

https://github.com/algolia/docsearch/blob/aa666decccb465a56d458d09379837342b520d73/packages/docsearch-react/src/StartScreen.tsx#L49

Which calls `<ul {...props.getListProps()}>` here:

https://github.com/algolia/docsearch/blob/7f670b6ba9d09bd8adf2455239c1c96ceb208851/packages/docsearch-react/src/Results.tsx#L31

Which gets its value from a different package:

https://github.com/algolia/autocomplete/blob/f05a701724d75fc3ade5e49c93d23077fb200371/packages/autocomplete-core/src/getPropGetters.ts#L302

And reads its value from here:

https://github.com/algolia/docsearch/blob/db8edeb818f8a788efe89b97f954308a1b1e2b32/packages/docsearch-react/src/DocSearchModal.tsx#L131

So it looks like `getListProps`  reads source only from the passed parameters. This PR passes the source over in the params.

Thanks for the review. Max.

### DIFF
--- a/cypress/e2e/search/actions.spec.ts
+++ b/cypress/e2e/search/actions.spec.ts
@@ -111,6 +111,10 @@ describe('Search', () => {
     cy.typeQueryNotMatching();
     cy.contains('No results for').should('be.visible');
   });
+
+  it('should not refer to Recent/Favorite in aria-controls', () => {
+    cy.get('.DocSearch-Input').should('not.have.attr', 'aria-controls');
+  });
 });
 
 describe('Recent and Favorites', () => {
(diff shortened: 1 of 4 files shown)


mxschmitt:
Status: Depends on a new version autocomplete - will rebase once https://github.com/algolia/autocomplete/pull/1302 has been released.

mxschmitt:
Should be ready for review @shortcuts.

mxschmitt:
@dylantientcheu PTAL - thanks!

dylantientcheu:
beautiful! thanks for handling this @mxschmitt 

fix the conflicts and let's merge it

mxschmitt:
rebased @dylantientcheu - thanks!

dylantientcheu (approved):
🚀

mxschmitt:
Just want to make sure that its not missing something - was the PR intentionally not merged yet? (trying to avoid more merge conflicts). Thanks.

8bittitan (approved):
Thanks for this @mxschmitt 🙌

mxschmitt:
@dylantientcheu when is the next release planned? Since our a11y team would like to verify that this fixes https://github.com/microsoft/playwright.dev/issues/1628.

PR #2555, opened 2025-04-05
feat(v4): add new UI

vascobettencourt:
### Description
- UI refresh.
- Mobile-friendly design.
- New default styling and support for CSS overrides.
- Updates to the search bar and its various states.


packages/docsearch-css/src/button.css
```diff
@@ -1,44 +1,45 @@
...
+}
+
 .DocSearch-Search-Icon {
+  color: var(--docsearch-highlight-color);
   stroke-width: 1.6;
 }
 
-.DocSearch-Button .DocSearch-Search-Icon {
+.DocSearch-Button {
   color: var(--docsearch-text-color);
```
8bittitan, line 34: Based on the designs, this should be updated to be `var(--docsearch-muted-color)` as well right?

vascobettencourt: Nice catch, I removed it

8bittitan (approved):
Nice work! Would like for 1 or 2 more reviews before merging just to make sure all is good 😄

NatanTechofNY (approved):

PR #2566, opened 2025-04-11
chore: Switch Preact to be a devDep of '@docsearch/js'

rschristian:
As `@docsearch/js` inlines Preact into its bundle, there's no reason for it to be listed as a dependency. Package managers will install Preact alongside this module but that'll never be used.

Won't have any effect on usage, just strips out an unnecessary dependency for those consuming from NPM (as opposed to a CDN, which tend to be a bit smarter about this sort of thing).


dylantientcheu (approved):
Sorry for the delay.

Nice catch @rschristian ! 👍🏽

PR #2574, opened 2025-04-17
feat(v4): ask-ai foundations :sparkles:

dylantientcheu:
This PR adds the foundation to implement AskAI on DocSearch:

- Check if ask ai is possible with the current config
- Add the AskAi context where conversations will live
- Keyword search should stay the same if the user is not using askAI.

We'll compare 2 very different approaches:

[`7f256e9` (#2574)](https://github.com/algolia/docsearch/pull/2574/commits/7f256e9442fd24b4181537132ee397ae33283d93) ask ai added as component. 

**pros:**
- completely separated from docsearch implementation
- easier and less code overhead

**cons:**
- doesn't integrate nicely with autocomplete.
- we lose autocomplete's accessibility
- duplicacy in code (styles, components etc)

---

[`fecb11a` (#2574)](https://github.com/algolia/docsearch/pull/2574/commits/fecb11a7bd613082d45b81785c37575090197c4f) ask ai as an autocomplete source

**pros:**
- better integration with autocomplete core
- fits in by providing nicer accessibility and keyboard navigation compared to the other approach

**cons:**
- much more code (some complex states have to be handled)
- any changes in autocomplete docsearch hits will have an effect on this


8bittitan (approved):
🛳️ ✨

NatanTechofNY:
Can we increase the size limit since it's failing

8bittitan (approved):
Awesome work! 👏

PR #2575, opened 2025-04-17
boosting with metadata

DanRoscigno:


DanRoscigno:
@kai687 thanks for your help working through this last year, i finally got around to submitting a PR to document this.

DanRoscigno:
Reviewers: I don't generally write JavaScript, so please let me know if I am not following best practices.

randombeeper:
Hi @DanRoscigno , nice contribution. I totally agree we need a bigger section on boosting, too many people are unaware of it, so it absolutely needs more visibility and explanation.

I was able to scan over the PR this afternoon but not dig too deeply on it. Two things jump out at me - the highlighting of the code seems to look fine for me in dark mode but not so much in light mode, not sure if I have something wrong on my end. 

The other thing is that I *think* this can be simplified by going this approach http://localhost:3000/docs/required-configuration/#introduce-global-information-as-meta-tags. Using slightly different meta tags should remove the need to have to modify the crawler. One down side to this approach is that IMO it is invisible and I could see easily confusing someone - "why do these results seem boosted when the crawler isn't doing anything special here?" I think it could be easy to miss. But we can potentially make that a separate section, as another way to accomplish the same result.

Will revisit again early next week.

DanRoscigno:
Thanks James, I will have a look at the highlighting. This is probably a change to your CSS.

Regarding the meta tags, if I am reading the doc correctly I could use:

```html
<meta name="docsearch:pageRank" content="100" />
```
That would be wonderful. I see that the [reactive docs](https://github.com/search?q=%3Cmeta+name%3D%22docsearch%3ApageRank%22&type=code) use this. I will give it a go and update the PR with this method. Probably I will put two scenarios in there, I do have a non-standard field that I extract.

randombeeper:
Yes, exactly, https://dataclient.io/docs and https://resthooks.io/docs are some examples of this.

I think it would be great to cover both approaches since some users will want to do one, or the other, for different potential reasons. But document whichever one(s) you want to do and I fill in the rest after the PR is merged.

DanRoscigno:
yup, I just finished testing `docsearch:pagerank` in the StarRocks docs and reverted the changes I made to my crawler, I will include both in the PR as I am extracting other data in our crawler.

DanRoscigno:
@randombeeper can you approve the netlify build so I can see the CSS issue, as I cannot get the docs to build on my end.

randombeeper:
Hi @DanRoscigno I think we got your docs working locally awhile back. Do you still want to try to finish this out?

PR #2603, opened 2025-05-23
fix: size and background scroll fixes

### LINKED ISSUE #2243: Search box is not visible in Safari iOS
## Description

The search box is not visible when opened after scrolling to the bottom of the page.
I noticed that it was displayed outside the screen at the top of the page
More, the overlay blocks scrolling back to the top of the page.

## Steps to reproduce

1. In safari iOS
2. Scroll bottom
3. Open search

Reproduction in sites :
- edge-components: https://edge-components.jrmc.dev/
- adonisJS: https://docs.adonisjs.com/guides/introduction
- edgeJS: https://edgejs.dev/docs/introduction
- Lucid orm: https://lucid.adonisjs.com/docs/introduction

## Environment

- OS: iOS 17.4.1
- Browser: Safari
- DocSearch version: 3.6.0

xav-ie:
Fixes #2243 

- **fix(docsearch-css): use correct size and position**
- **fix(docksearch-css): prevent mobile body scrolling**

Makes the modal sized correctly and prevents background scroll on mobile (the trade-off of no-background-scroll is that you get scrolled to top of page, but this is only necessary on mobile).

---

-webkit-fill-available is no longer even an option:
- https://caniuse.com/intrinsic-width
- https://developer.mozilla.org/en-US/docs/Web/CSS/WebKit_Extensions

`max-content` is close equivalent, but it is not what you want when you actually want the thing to take up the full viewport.

### DIFF
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -576,10 +576,8 @@ assistive tech users */
   }
 
   .DocSearch-Container {
-    height: 100vh;
-    height: -webkit-fill-available;
     height: calc(var(--docsearch-vh, 1vh) * 100);
-    position: absolute;
+    height: 100dvh;
   }
 
   .DocSearch-Footer {
@@ -597,9 +595,8 @@ assistive tech users */
   .DocSearch-Modal {
     border-radius: 0;
     box-shadow: none;
-    height: 100vh;
-    height: -webkit-fill-available;
     height: calc(var(--docsearch-vh, 1vh) * 100);
+    height: 100dvh;
     margin: 0;
     max-width: 100%;
     width: 100%;
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -570,6 +570,12 @@ assistive tech users */
     --docsearch-spacing: 10px;
     --docsearch-footer-height: 40px;
   }
+  /* Prevent body scroll on modal-open for mobile */
+  /* https://stackoverflow.com/a/24727206/6276948 */
+  body:has(.DocSearch-Container) {
+    overflow: hidden;
+    position: fixed;
+  }
 
   .DocSearch-Dropdown {
     height: 100%;


packages/docsearch-css/src/modal.css
```diff
@@ -576,10 +585,9 @@ assistive tech users */
   }
 
   .DocSearch-Container {
-    height: 100vh;
```
xav-ie, line 579: this line is not necessary. 

`vh` feature/unit came after `calc`. In your `calc` function, you use `1vh * 100` as a fallback. This means this line is redundant.

packages/docsearch-css/src/modal.css
```diff
@@ -576,10 +585,9 @@ assistive tech users */
...
-    height: -webkit-fill-available;
+    height: max-content;
```
xav-ie, line 588: I would remove this... but wanted to keep changes minimal. `-webkit-fill-available` should not be used when there is `dvh` that works as expected and is widely supported. `max-content` seems to be the successor...

packages/docsearch-css/src/modal.css
```diff
@@ -1,5 +1,14 @@
 /* Body modifier */
 
+/* Prevent body scroll on modal-open for mobile */
+/* https://stackoverflow.com/a/24727206/6276948 */
+@media (max-width: 768px) {
+  body:has(.DocSearch-Container) {
+    overflow: hidden;
+    position: fixed;
+  }
```
benface, line 9: Shouldn't this be colocated with the other mobile styles, instead of repeating `@media (max-width: 768px)`?

xav-ie: Fair. Fixed.

packages/docsearch-css/src/modal.css
```diff
@@ -597,9 +602,9 @@ assistive tech users */
   .DocSearch-Modal {
     border-radius: 0;
     box-shadow: none;
-    height: 100vh;
-    height: -webkit-fill-available;
+    height: max-content;
```
benface, line 605: Does any browser support `max-content` but not `calc()`? I doubt it, so I would remove this line as it will always be overridden by the next one.

packages/docsearch-css/src/modal.css
benface: Does any browser support max-content but not calc()? I doubt it, so I would remove this line as it will always be overridden by the next one.

benface (approved):
I don't work at Algolia, but since you asked for my review, LGTM :)

packages/docsearch-css/src/modal.css
xav-ie: Yes, I think that is also good point. Thank you. Fixed!

dylantientcheu (approved):
looking good. 💯 
thanks for working on this

8bittitan (approved):
Thank you for this!

PR #2675, opened 2025-07-21
chore(deps): change react to dev deps on js package

dylantientcheu:
Since the js package is shipped as a bundle, there's no need to bundle the react dependency in it. 

We changed it to a dev dependency. It should help reduce the bundle size and fix a few issues with installing the dependency in non-react frameworks.

see https://github.com/vuejs/vitepress/issues/472 for more context on this fix.

### DIFF
--- a/packages/docsearch-js/package.json
+++ b/packages/docsearch-js/package.json
@@ -32,10 +32,8 @@
     "on:change": "yarn build",
     "watch": "nodemon --watch src --exec \"yarn on:change\" --ignore dist/ --ext ts,tsx"
   },
-  "dependencies": {
-    "@docsearch/react": "4.0.0-beta.2"
-  },
   "devDependencies": {
+    "@docsearch/react": "4.0.0-beta.2",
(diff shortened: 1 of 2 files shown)


dylantientcheu:
@shortcuts IMO both esm and umd builds from the `docsearch-js` package do not need a React/Preact environment at runtime.

They are both shipped as pureJS bundle with no need for external packages:
see ESM: https://cdn.jsdelivr.net/npm/@docsearch/js@beta/dist/esm/index.js
and UMD: https://cdn.jsdelivr.net/npm/@docsearch/js@beta/dist/umd/index.js

They are used during build, but not needed when shipped, there's therefore no need to keep the package as direct dependency.

brc-dd:
Hi, vitepress maintainer here. I just saw the reverted commit. I think I didn't convey clearly.

`@docsearch/react` shouldn't be externalized. Because you need to replace react in it with preact. Externalizing will break stuff.

I was saying that keep it internal, but move it to dev deps. So other packages won't need to install it as it is unused.

PR #2705, opened 2025-08-07
fix: demos-transformitems

dylantientcheu:


examples/demo-react/src/examples/w-hit-transformItems.tsx
```diff
@@ -0,0 +1,61 @@
...
+            lvl4: null,
+            lvl5: null,
+            lvl6: null,
+          },
+          url_without_anchor: item.domain + item.path,
+          type: 'content' as const,
+          anchor: null,
+          _highlightResult: item._highlightResult,
+          _snippetResult: item._snippetResult,
+        }));
```
NatanTechofNY, line 56: Think we can add this functionality to the playground?

dylantientcheu: We unfortunately can't because these are functions and objects. Not just strings.

NatanTechofNY (approved):

PR #2716, opened 2025-08-13
fix(docsearch): localstorage quote handling

NatanTechofNY:


packages/docsearch-react/src/utils/storage.ts
```diff
@@ -82,7 +166,23 @@ export function createObjectStorage<TItem>(key: string) {
...
+              cleanupDocSearchStorage();
+              // Retry the operation
+              window.localStorage.setItem(key, JSON.stringify(item));
+            } catch {
+              // If still failing, silently fail to prevent crashes
+            }
+          } else {
+            // For other localStorage errors, silently fail to prevent crashes
+          }
+        }
```
dylantientcheu, line 185: since it is the same as above, we can extract it in a function, put this in this same file.

packages/docsearch-react/src/utils/storage.ts
```diff
@@ -1,3 +1,71 @@
...
+    if (key.includes('__DOCSEARCH_')) {
+      const value = window.localStorage[key];
+      docSearchKeys.push({ key, size: value.length + key.length });
+    }
+  }
+
+  // Sort by size (largest first) to remove the most impactful items
+  docSearchKeys.sort((a, b) => b.size - a.size);
+
+  // Remove up to half of the DocSearch items, starting with the largest
```
dylantientcheu, line 40: this will for sure make users lose conversations 🤷🏽.
it sounds okay for docsearch, but for other usages than that, it might not be the best thing.

dylantientcheu: A potential long term solution would be to migrate all of this to indexedDB which has a much higher limit, but is quite complex to work with.

I'd say the current solution works. We'll keep observing

dylantientcheu (commented):
Nice PR. I'm approving but,
I left a few comments. Code structure and feature

packages/docsearch-react/src/utils/storage.ts
NatanTechofNY: I agree

dylantientcheu (approved):

PR #2736, opened 2025-08-27
feat(v4): Add support for multi-index search

8bittitan:
## What
Adds support for multi-index search during keyword search. This does not impact AskAI in any way.

- Add new `indices` property to accept a list of indexes for search
- Add deprecation messaging to `indexName` and `searchParameters` properties
- Manage backwards compatibility for now until `indexName` and `searchParameters` are fully removed
- Add another example of multi-index search

## Docs
- Adds documentation for `indices` and deprecation warnings for `indexName` and `searchParameters`
- Enables `respectPrefersColorScheme` within Docusaurus config

indices_js_docs

indices_react_docs

index_name_deprecation

search_parameters_deprecation

## Dependencies
- [x] Add documentation around new and deprecated properties

### DIFF
--- a/packages/docsearch-react/src/DocSearch.tsx
+++ b/packages/docsearch-react/src/DocSearch.tsx
@@ -72,7 +72,9 @@ export interface DocSearchProps {
    */
   indexName?: string;
   /**
-   * List of indices to be used for search.
+   * List of indices and _optional_ searchParameters to be used for search.
+   *
+   * @see {@link https://docsearch.algolia.com/docs/api#indices}
    */
   indices?: Array<DocSearchIndex | string>;
   /**
--- a/packages/docsearch-react/src/DocSearch.tsx
+++ b/packages/docsearch-react/src/DocSearch.tsx
@@ -51,6 +51,11 @@ export type DocSearchAskAi = {
   };
 };
 
+export interface DocSearchIndex {
+  name: string;
+  searchParameters?: SearchParamsObject;
+}
+
 export interface DocSearchProps {
   /**
    * Algolia application id used by the search client.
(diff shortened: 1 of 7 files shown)


pipeline1987 (approved):

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -214,6 +249,7 @@ export function DocSearch({ ...props }: DocSearchProps): JSX.Element {
             translations={props?.translations?.modal}
             isAskAiActive={isAskAiActive}
             canHandleAskAi={canHandleAskAi}
+            indexes={indexes}
```
slorber, line 252: Why is DocSearch prop named `indices`, while subcomponents accept an `indexes` prop? Isn't that a typo? Why not always a consistent name?

packages/docsearch-react/src/__tests__/api.test.tsx
```diff
@@ -350,4 +350,18 @@ describe('api', () => {
...
+    });
+
+    it('renders with new indices prop', () => {
+      expect(() => render(<DocSearch indexName={undefined} indices={[{ name: 'testing' }]} />)).not.toThrowError();
+    });
+
+    it('throws an error if indexName or indices are not provided', () => {
+      expect(() => render(<DocSearch indexName={undefined} indices={[]} />)).toThrowError();
+    });
+  });
```
slorber, line 366: Those tests only cover the root API, but do not cover `<DocSearchButton>` and `<DocSearchModal>` APIs

slorber (commented):
Reporting a retro-compatibility issue we got after upgrading from beta.8 to GA

Please tell us if you plan to restore API retro-compatibility for subcomponents too, or if we now have to handle this retro-compatibility ourselves.

PR #2740, opened 2025-08-28
[v4]Add keyboardShortcuts property to allow users to customize shortcuts

GregBrimble:


8bittitan (approved):
This is great! Thanks especially for adding tests and docs 🙌 .

Adding approval, but would also like to have @dylantientcheu take a look as well.

dylantientcheu (approved):
Thanks for this PR.

We'll probably allow more customization for the shortcuts in the future, but we are yet to decide the API.

PR #2741, opened 2025-08-28
[v3] Add `keyboardShortcuts` property to allow users to customize shortcuts

GregBrimble:
Hello again!

Not sure if you're accepting features on v3 any more, but here is a backport of #2740, in case you did want to merge it into v3 too.

Thank you!

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -6,11 +6,11 @@
     },
     {
       "path": "packages/docsearch-react/dist/umd/index.js",
-      "maxSize": "27 kB"
+      "maxSize": "27.5 kB"
     },
     {
       "path": "packages/docsearch-js/dist/umd/index.js",
-      "maxSize": "35.5 kB"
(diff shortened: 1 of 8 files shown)


packages/website/docs/api.mdx
8bittitan, line 1: @GregBrimble Could you please add these docs changes to `packages/website/versioned_docs/version-v3/api.mdx` within #2740 ? Or even a different PR since #2740 is already approved 😄

8bittitan (changes requested):
@GregBrimble Totally fine with this backport, only change would be to not make the docs changes in this PR since our website versioning doesn't work the same as package level versioning.

Also feel free to bump the `bundlesize.config.json` sizes for CI 😄

packages/website/docs/api.mdx
GregBrimble: https://github.com/algolia/docsearch/pull/2742

PR #2746, opened 2025-09-04
feat(askai): ai-sdk v5

8bittitan:
# What
Updates to use the `ai-sdk` V5. With it came minor changes to the streaming data response, types were updated, and new way of controlling the transport layer of the request.

## Preview (using gpt-5-mini)

https://github.com/user-attachments/assets/9ebbcd00-9d0c-4031-8b50-461c94d66972

### DIFF
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -797,7 +797,10 @@ assistive tech users */
   background: none;
   margin-left: 4px;
   text-overflow: ellipsis;
-  text-decoration: none !important;
+}
+
+.DocSearch-Hit-AskAIButton-title mark {
+  text-decoration: none;
 }
 
 @keyframes fade-in {


NatanTechofNY (approved):

packages/docsearch-css/src/modal.css
```diff
@@ -798,6 +797,7 @@ assistive tech users */
   background: none;
   margin-left: 4px;
   text-overflow: ellipsis;
+  text-decoration: none !important;
```
dylantientcheu, line 800: can we do this without using `!important`? most frameworks will not like it.

packages/docsearch-react/src/AskAiScreen.tsx
```diff
@@ -120,127 +126,113 @@ function AskAiExchangeCard({
...
-                      }
-                    }
-                    // fallback for unknown tool, should never happen in theory. :shrug:
+                  case 'output-available':
                     return (
-                      <span key={index} className="text-sm italic shimmer">
-                        {translations.thinkingText || 'Thinking...'}
-                      </span>
+                      <div key={index} className="DocSearch-AskAiScreen-MessageContent-Tool Tool--Result">
+                        <SearchIcon />
```
dylantientcheu, line 201: thanks for removing aggregated tool calling, it is not so useful and sometimes doesn't work

packages/docsearch-css/src/modal.css
NatanTechofNY: We tried without it and it didn't respect it. I guess @8bittitan can try again

PR #2764, opened 2025-09-16
feat: Split docsearch-react package exports

8bittitan:
Explicitly exports `DocSearchModal`, `DocSearchButton`, `useDocSearchKeyboardEvents`, `useTheme` and `version`  as their own standalone exports. This is to reduce overall bundle size if wanting to use these components on their own.

With this you could now import like the following:

`import { DocSearchButton } from '@docsearch/react/button'`
`import { DocSearchModal } from '@docsearch/react/modal'`
`import { useDocSearchKeyboardEvents } from '@docsearch/react/useDocSearchKeyboardEvents'`
`import { useTheme } from '@docsearch/react/useTheme'`
`import { version } from '@docsearch/react/version'`

The new `@docsearch/react` `dist` directory will look like the following:

```
.
├── esm
│   ├── DocSearchButton.d.ts
│   ├── DocSearchButton.js
│   ├── DocSearchModal.d.ts
│   ├── DocSearchModal.js
│   ├── index.d.ts
│   ├── index.js
│   ├── useDocSearchKeyboardEvents.d.ts
│   ├── useDocSearchKeyboardEvents.js
│   ├── useTheme.d.ts
│   ├── useTheme.js
│   ├── version.d.ts
│   └── version.js
└── umd
    ├── DocsearchButton.js
    ├── DocsearchButton.js.map
    ├── DocSearchModal.js
    ├── DocSearchModal.js.map
    ├── index.js
    ├── index.js.map
    ├── useDocSearchKeyboardEvents.js
    ├── useDocSearchKeyboardEvents.js.map
    ├── useTheme.js
    ├── useTheme.js.map
    ├── version.js
    └── version.js.map
```

PR #2766, opened 2025-09-18
Example of multiple entrypoints

slorber:
Doing such a thing can help library consumers to better code-split the code.

I tested on Docusaurus, and we could significantly minimize the impact of the modal + askAi on bundle size (saving 400kb) if you exposed 3 entrypoints:
- `/button` 
- `/version` 
- `/useDocSearchKeyboardEvents` 

Before:
> -rw-r--r--@ 1 sebastienlorber  staff  1167928 Sep 18 18:33 website/build/assets/js/main.2c9e14df.js

After: 
> -rw-r--r--@ 1 sebastienlorber  staff  735434 Sep 18 18:34 website/build/assets/js/main.029ed3dc.js

Note: it could be beneficial to expose a `/modal` entrypoint too, otherwise we'd load all the button/useDocSearchKeyboardEvents twice, but the most important is to be able to use the button with lightweight code, and code-split/lazy-load the heavy modal

### DIFF
--- a/packages/docsearch-react/button.js
+++ b/packages/docsearch-react/button.js
@@ -1,1 +1,1 @@
-export { DocSearchButton } from './dist/esm';
+export { DocSearchButton } from './dist/esm/button.js';
--- a/packages/docsearch-react/rollup.config.js
+++ b/packages/docsearch-react/rollup.config.js
@@ -36,6 +36,32 @@ export default [
     ],
   },
   {
+    input: 'src/button.ts',
+    external: ['react', 'react-dom'],
+    output: [
+      {
+        globals: {
+          react: 'React',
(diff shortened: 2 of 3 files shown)


8bittitan:
@slorber Thanks for this! I do have an open PR that tackles this as well: #2764 which also handles the modal. Does that look like something that would also lighten the bundle size issue?

slorber:
Thanks @8bittitan 

Yes that would be helpful, so I'm closing this PR.

However, we also use the following imports:

```tsx
import {version, useDocSearchKeyboardEvents} from '@docsearch/react';
```

We'd also need these to be exported independently, or we'd have to stop using these exports and find alternatives

PR #2768, opened 2025-09-19
feat: curate AskAI search parameters

vascobettencourt:


packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -29,6 +29,16 @@ export type DocSearchTransformClient = {
   transporter: Pick<LiteClient['transporter'], 'algoliaAgent'>;
 };
 
+// Define the specific search parameters allowed for AskAI
+export type AskAiSearchParameters = {
+  facetFilters?: string[];
+  filters?: string;
+  attributesToRetrieve?: string[];
+  attributesToSnippet?: string[];
```
NatanTechofNY, line 37: Not passable

PR #2812, opened 2025-12-02
feat(askai): add thread depth limit logic

vascobettencourt:
## What's New

Adds frontend support for the AI-217 (Thread Depth Exceeded) error. When a conversation reaches its maximum depth limit, users now see a clear error message with an option to start a new conversation.

## Implementation

When the backend returns an AI-217 error:
- A pink error banner appears below the search input
- The input is disabled with "Conversation limit reached" placeholder
- Users can click "Start a new conversation" to continue
- The failed message is automatically hidden from the conversation

## Bundle Impact

+0.48 KB (0.42% increase)

https://github.com/user-attachments/assets/2a1f04b1-0f73-4163-a96e-fac16681b4a1

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,6 +3,6 @@ YARN_VERSION = "1.22.5"
 
 [[redirects]]
 from = "/apply/"
-to = "https://dashboard.algolia.com/users/sign_up?selected_plan=docsearch"
+to = "https://dashboard.algolia.com/users/sign_up?selected_plan=docsearch&utm_source=docsearch.algolia.com&utm_medium=referral&utm_campaign=docsearch&utm_content=apply"
 status = 301
 force = true
--- a/packages/website/docs/docsearch-program.md
+++ b/packages/website/docs/docsearch-program.md
@@ -88,7 +88,7 @@ Please reach out to our [support](https://algolia.com/support) team.
 
 ## Can I have multiple projects under the same Algolia application?
 
-We recommend having a single Algolia application per project. Please [apply](https://dashboard.algolia.com/users/sign_up?selected_plan=docsearch) if you'd like to use DocSearch in an other project of yours.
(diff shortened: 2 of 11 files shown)


packages/docsearch-react/src/AskAiScreen.tsx
```diff
@@ -52,6 +52,14 @@ export type AskAiScreenTranslations = Partial<{
    * Error title shown if there is an error while chatting.
    */
   errorTitleText: string;
+  /**
+   * Message shown when thread depth limit is exceeded (AI-217 error).
+   */
+  threadDepthExceededMessage: string;
```
8bittitan: If we move rendering of the error to the `AskAiScreen` then we should keep and use these. If we can't move it for some reason, we should remove these.

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -884,6 +899,21 @@ export function DocSearchModal({
           />
         </header>
 
+        {/* Thread Depth Error */}
+        {showThreadDepthError && (
```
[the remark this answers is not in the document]
vascobettencourt: done!

packages/docsearch-react/src/SearchBox.tsx
```diff
@@ -116,12 +118,20 @@ export function SearchBox({
...
+  const isThreadDepthError = props.isThreadDepthError || false;
   let searchPlaceholder = props.placeholder;
 
   if (askAiState === 'new-conversation') {
     searchPlaceholder = newConversationPlaceholder;
   }
 
+  // Override placeholder when thread depth error occurs (only in Ask AI mode)
+  if (isThreadDepthError && props.isAskAiActive) {
+    searchPlaceholder = 'Conversation limit reached';
```
8bittitan: Sorry I missed this earlier, but might as well add this as a translation too if you could!

PR #2832, opened 2026-01-08
feat(askai): Allow use of Agent Studio for chat backend

8bittitan:
## Summary

- Add experimental agentStudio prop to enable using Algolia's Agent Studio as an alternative chat backend
- Hide feedback (thumbs up/down) buttons when agentStudio is enabled since Agent Studio doesn't support the feedback API
- Add documentation for the new agentStudio parameter

## Changes

New Feature: Agent Studio Backend Support
- Added agentStudio boolean prop to DocSearch, DocSearchModal, and Sidepanel components
- When enabled, the Ask AI chat uses Algolia's Agent Studio transport instead of the default Ask AI backend
- Created new getAgentStudioTransport function in useAskAi.ts to handle the different API endpoint
Feedback Buttons Removal for Agent Studio
- Feedback buttons are now conditionally hidden when agentStudio={true} in both:
  - DocSearchModal via AskAiScreenFooterActions
  - Sidepanel via ConversationActions
- The Copy button remains visible regardless of the backend used

## Documentation
- Added agentStudio API documentation to both api.mdx and sidepanel/api-reference.mdx
- Marked as experimental, targeted for stable release in v5.0.0

## Demo
- Added Agent Studio example in examples/demo-react/src/examples/agent-studio.tsx

### DIFF
--- a/packages/website/docs/v4/askai-errors.mdx
+++ b/packages/website/docs/v4/askai-errors.mdx
@@ -60,7 +60,7 @@ These error codes represent issues with the AI stream or upstream LLM providers.
 
 ### AI-208 - Provider API call error {#ai-208}
 
-There was an issue communicating with the upstream provider’s API. This can occur when the provider returns an unexpected response or a timeout.
+There was an issue communicating with the upstream provider's API. This can occur when the provider returns an unexpected response or a timeout.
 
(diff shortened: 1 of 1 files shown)


NatanTechofNY (approved):

dylantientcheu (commented):
Few nits.
But ready to go!

PR #2869, opened 2026-03-05
chore: deps upgrade

dylantientcheu:
Fixing deps with docsearch.

https://github.com/algolia/docsearch/security/dependabot


8bittitan:
@dylantientcheu Is this still needed? Maybe we should close and open a new one now after `5.0.0`?

PR #2877, opened 2026-04-07
feat(agentStudio): surface thread depth error

pipeline1987:
## 🤔 What

- Agent Studio / thread-depth handling thread-depth UX.
- Stable chat sessions, and session reset fixes.

## How 🧐

- Surfaces thread-depth errors in the modal and sidepanel with clearer copy and a path to start a new conversation.
- Sidepanel: thread-depth banner sits above the prompt; conversation list styling updated in sidepanel.css.
- Modal: small layout tweaks on the Ask AI error block for consistency.

## 🧪 How to test

- On Ask AI, hit thread-depth limit → see error + start new conversation; confirm the next turn uses a new API thread, not the capped one.
- In the modal, send several follow-ups (Enter) → same conversation until you explicitly start new or leave Ask AI.

## 📸 Screenshots

image

image

### DIFF
--- a/packages/docsearch-css/src/sidepanel.css
+++ b/packages/docsearch-css/src/sidepanel.css
@@ -661,6 +661,11 @@ html[data-theme="dark"] .DocSearch-Sidepanel-Prompt--stop:hover {
   margin: 0;
 }
 
+.DocSearch-Sidepanel .DocSearch-Sidepanel-ThreadDepthBanner-apiMessage {
+  font-weight: 700;
+  margin-bottom: 0.5rem;
+}
+
 @media screen and (min-width: 769px) {
   .DocSearch-Sidepanel-ConversationScreen {
     padding: 1rem 0;
--- a/packages/docsearch-css/src/sidepanel.css
+++ b/packages/docsearch-css/src/sidepanel.css
@@ -19,12 +19,16 @@
   --docsearch-sidepanel-hit-highlight-color: var(--docsearch-hit-highlight-color);
   --docsearch-sidepanel-button-background: var(--docsearch-sidepanel-background);
   --docsearch-sidepanel-button-background-dark: var(--docsearch-sidepanel-background);
+  --docsearch-sidepanel-thread-depth-banner-bg: #fff1f2;
+  --docsearch-sidepanel-thread-depth-banner-border: #fecdd3;
 }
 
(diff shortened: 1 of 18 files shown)


dylantientcheu (approved):

packages/docsearch-css/src/sidepanel.css
```diff
@@ -19,12 +19,16 @@
   --docsearch-sidepanel-hit-highlight-color: var(--docsearch-hit-highlight-color);
   --docsearch-sidepanel-button-background: var(--docsearch-sidepanel-background);
   --docsearch-sidepanel-button-background-dark: var(--docsearch-sidepanel-background);
+  --docsearch-sidepanel-thread-depth-banner-bg: #fff1f2;
```
dylantientcheu, line 22: Can we use an already existing color ?

dylantientcheu (changes requested):

packages/docsearch-react/src/Sidepanel/Sidepanel.tsx
```diff
@@ -406,9 +455,27 @@ function SidepanelInner(
...
+                    className="DocSearch-ThreadDepthError-Link"
+                    onClick={handleStartNewConversation}
+                  >
+                    {startNewConversationButtonText}
+                  </button>{' '}
+                  to continue.
+                </p>
+              </div>
+            ) : null
+          }
```
LorrisSaintGenez: IMO, `threadDepthBanner` should always be set, we conditionally render it on `PromptForm` instead of here, based on `isThreadDepthError`, so each component is responsible for rendering its props

LorrisSaintGenez (changes requested):
👏 One comment

packages/docsearch-react/src/Sidepanel/Sidepanel.tsx
pipeline1987: Yeah, it makes sense 👍🏽

packages/docsearch-css/src/sidepanel.css
pipeline1987: I tried, but didn't found a similar color :/

LorrisSaintGenez (approved):

dylantientcheu (approved):

PR #2888, opened 2026-05-29
feat(askai): add Agent Studio memory support

8bittitan:
## Why

Agent Studio can now persist and recall user-specific context across conversations through its memory feature. Without support in DocSearch, these memory operations were invisible to users, the agent would silently store and retrieve memories with no feedback, and there was no way to scope memory to an authenticated user.

This PR makes the memory feature first-class in the Ask AI experience so that:

- **Users get feedback when memory is used.** Memory tool calls (`algolia_ponder`, `algolia_memorize`, `algolia_memory_search`) now surface clear messaging ("Saved to memory" / "Used memory to enhance results") instead of being dropped as unrecognized tool calls. This is opt-in via `memory.enabled` to avoid changing behavior for integrations that don't use it.
- **Memory can be tied to a specific user.** A `memory.userToken` (JWT) is forwarded as the `x-algolia-secure-user-token` header so the agent reads/writes the correct user's memory, enabling per-user personalization without leaking memory across users.

## Notes

- Translations for the new memory messages are exposed and overridable for i18n.
- Custom tool renderers still take precedence over the built-in memory rendering, preserving existing override behavior.
- The shared `ToolState` primitive was extracted so the new `MemoryTool` and existing `SearchTool` render consistently.

## Testing

- Enable memory support for your agent on the Algolia Dashboard ([docs](https://www.algolia.com/doc/guides/algolia-ai/agent-studio/how-to/memory/overview#enable-memory))
- Enable showing the memory tool calls on the DocSearch UI `memory.enabled = true`
- Use something like [jwt.io's](https://www.jwt.io/) ([Agent Studio docs](https://www.algolia.com/doc/guides/algolia-ai/agent-studio/how-to/user-authentication#set-up-user-authentication)) encoder to generate a JWT to use and pass it through props `memory.userToken = JWT`

## Screenshots
Screenshot 2026-05-29 at 12 59 13 PM

### DIFF
--- a/packages/docsearch-react/src/useAskAi.ts
+++ b/packages/docsearch-react/src/useAskAi.ts
@@ -18,7 +18,7 @@ import { createStoredConversations } from './stored-searches';
 import { type AIMessage, type ToolCalls } from './types/AskiAi';
 import { EMPTY_TOOLS } from './utils/ai';
 
-import type { AgentStudioSearchParameters, AskAiSearchParameters, StoredAskAiState } from '.';
+import type { AgentStudioSearchParameters, AskAiSearchParameters, Memory, StoredAskAiState } from '.';
 
 type UseChat = UseChatHelpers<AIMessage>;
 
@@ -31,6 +31,7 @@ type UseAskAiParams = {
   searchParameters?: AskAiSearchParameters;
   tools: ToolCalls;
   agentStudio: boolean;
+  memory?: Memory;
 } & (
   | {
       agentStudio: false;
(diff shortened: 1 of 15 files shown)


pipeline1987 (approved):

NatanTechofNY (approved):

packages/docsearch-react/src/useAskAi.ts
```diff
@@ -134,6 +139,7 @@ export const useAskAi: UseAskAi = ({
             appId,
             assistantId: assistantId ?? '',
             searchParameters,
+            userToken: memory?.userToken,
```
LorrisSaintGenez, line 142: Shouldn't we check too that `memory` is enabled here too before sending the userToken?

8bittitan: No, `memory.enabled` is used to determine if we are showing the tool calls or not.

PR #2891, opened 2026-06-01
feat(askai): Aggregate MCP search tool calls

8bittitan:
## Why

When Ask AI runs searches through the Algolia MCP server, each query comes back as a separate `algolia_search_index` (or `algolia_search_index_*`) tool call. Previously, only the native `searchIndex` tool was recognized for aggregation, so MCP-backed searches were rendered as a noisy list of individual tool-call entries instead of being collapsed into a single grouped result. This made multi-query MCP searches harder to scan and inconsistent with how native search is presented.

This change teaches the grouping logic to treat MCP search tools as first-class search results so consecutive searches — native or MCP, or a mix of both — are aggregated into one clean entry.

## What this enables

- Consecutive MCP search calls collapse into a single aggregated result, matching native `searchIndex` behavior.
- Mixed native + MCP search sequences group together rather than fragmenting.
- Query extraction accounts for the difference between the two tool shapes (native exposes the query on its `output`, MCP on its `input`), and empty/whitespace queries are skipped.
- Custom tools that merely share the `algolia_search_index` prefix (e.g. `algolia_search_indexer`) are intentionally excluded to avoid false grouping.

## Notes

- Shared type guards (`isSearchToolPart`, `isSearchOutputPart`, etc.) were centralized in `utils/ai.ts` and removed from `ToolCall.tsx` to keep the search-tool detection logic in one place.
- New unit tests cover the MCP guard and the grouping behavior, including the prefix edge cases and empty-query handling.
- `AGENTS.md` was updated to reflect the current toolchain (Playwright instead of Cypress, `--run` for tests).

## Dependencies
- [x] Point target branch to `v5` after #2890 has been merged


packages/docsearch-react/src/utils/ai.ts
```diff
@@ -113,3 +123,27 @@ export const EMPTY_TOOLS: Readonly<ToolCalls> = Object.freeze({});
...
+  );
+}
+
+export function isSearchIndexOutputPart(part: AIMessagePart): part is SearchIndexOutputPart {
+  return part.type === 'tool-searchIndex' && part.state === 'output-available';
+}
+
+export function isAlgoliaMCPSearchOutputPart(part: AIMessagePart): part is AlgoliaMCPSearchOutputPart {
+  return (
+    isAIToolPart(part) &&
```
LorrisSaintGenez, line 141: We don't need `isAIToolPart(part)` here, since the following line checks the same property and more restrictively

packages/docsearch-react/src/utils/ai.ts
```diff
@@ -113,3 +123,27 @@ export const EMPTY_TOOLS: Readonly<ToolCalls> = Object.freeze({});
...
+    (part.type === 'tool-algolia_search_index' || part.type.startsWith('tool-algolia_search_index_')) &&
+    part.state === 'output-available'
+  );
+}
+
+export function isSearchOutputPart(part: AIMessagePart): part is SearchOutputPart {
+  return isAIToolPart(part) && isSearchToolPart(part) && part.state === 'output-available';
```
LorrisSaintGenez, line 148: Same here 👀 `isSearchToolPart` > `isAIToolPart`

LorrisSaintGenez (approved):
Non-blocking comment

packages/docsearch-react/src/utils/ai.ts
8bittitan: Both cases of this are more for narrowing the type of `part`. Otherwise the `part.state` doesn't know what it's supposed to be 🤔

NatanTechofNY (approved):

packages/docsearch-react/src/utils/ai.ts
LorrisSaintGenez: Ah right 😮‍💨

galfa9120-hash:
> ## Why
> When Ask AI runs searches through the Algolia MCP server, each query comes back as a separate `algolia_search_index` (or `algolia_search_index_*`) tool call. Previously, only the native `searchIndex` tool was recognized for aggregation, so MCP-backed searches were rendered as a noisy list of individual tool-call entries instead of being collapsed into a single grouped result. This made multi-query MCP searches harder to scan and inconsistent with how native search is presented.
> 
> This change teaches the grouping logic to treat MCP search tools as first-class search results so consecutive searches — native or MCP, or a mix of both — are aggregated into one clean entry.
> 
> ## What this enables
> * Consecutive MCP search calls collapse into a single aggregated result, matching native `searchIndex` behavior.
> * Mixed native + MCP search sequences group together rather than fragmenting.
> * Query extraction accounts for the difference between the two tool shapes (native exposes the query on its `output`, MCP on its `input`), and empty/whitespace queries are skipped.
> * Custom tools that merely share the `algolia_search_index` prefix (e.g. `algolia_search_indexer`) are intentionally excluded to avoid false grouping.
> 
> ## Notes
> * Shared type guards (`isSearchToolPart`, `isSearchOutputPart`, etc.) were centralized in `utils/ai.ts` and removed from `ToolCall.tsx` to keep the search-tool detection logic in one place.
> * New unit tests cover the MCP guard and the grouping behavior, including the prefix edge cases and empty-query handling.
> * `AGENTS.md` was updated to reflect the current toolchain (Playwright instead of Cypress, `--run` for tests).
> 
> ## Dependencies
> * [x]  Point target branch to `v5` after [feat(askai): Feedback notes and tags #2890](https://github.com/algolia/docsearch/pull/2890) has been merged

Open play

PR #2896, opened 2026-06-09
feat(v5): UI updates

8bittitan:
### What

This PR is primarily a visual refresh of the DocSearch modal for v5, with a few supporting structural changes to make the new UI maintainable. It is mostly presentation-focused — there are no significant behavioral changes to search itself.

### Why

The existing modal styling had drifted from the intended v5 design, and several patterns (ad-hoc panels, manually built source lists, index-based collection lookups) made the UI hard to evolve and prone to breakage. This PR aligns the look-and-feel with the target design and replaces those fragile patterns with reusable, intention-revealing primitives.

### Key changes and rationale

- **Refreshed light & dark themes.** Theme variables were reworked (true white
  modal background, dedicated focus-background tokens, larger radius/spacing,
  new popover tokens) so the two themes stay visually consistent and are easier
  to tune from one place rather than scattered overrides.

- **Sources moved into a popover.** The Ask AI "related sources" list was a
  bespoke inline panel. It's now a `SourcesPanel` built on a shared `Popover`
  primitive (`@base-ui/react`), which keeps sources out of the answer flow,
  handles positioning/scroll-out-of-view, and gives us a reusable popover for
  future UI. This is the reason for the new `@base-ui/react` dependency.

- **Collection lookups by id instead of index.** Components previously reached
  into `state.collections[2]` etc., which silently breaks whenever source
  ordering changes. A small `getCollection`/`SOURCE_IDS` helper makes these
  lookups explicit and order-independent. `createDocSearchSources` was updated
  to use the shared ids and to emit a single hits source per index (instead of
  one per group), simplifying rendering.

- **Extracted shared hooks.** Platform-key detection (`usePlatformKeys`) was
  duplicated inline; it's now a hook reused by the button and search box. A new
  `useRelativeFormattedDate` hook renders human-friendly timestamps for recent
  Ask AI conversations.

- **Search box affordances.** Added an explicit "Ask AI" action button and
  cmd/ctrl+Enter shortcut so users can jump to Ask AI from any query, plus a
  clear-button divider, surfacing actions that were previously hidden.

- **Simplified results rendering.** Result rows, section headers, and the Ask AI
  entry were restructured to match the new design (consistent section titles,
  source icons, dropped the tree-connector SVG) for a cleaner, more uniform list.

### Screenshots

| Light mode | Dark mode |
|---|---|
| v5_light |  v5_dark |
| v5_results_light | v5_results_dark |
| v5_chat_light | v5_chat_dark |
| v5_sources_light | v5_sources_dark |

### Notes

- Test files were moved alongside their components/hooks and updated accordingly.
- E2E selectors and bundle-size limits were adjusted to match the new markup/CSS.
- Added theme toggle and dark theme to react demo example app

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -180,6 +180,7 @@
       "version": "4.6.0",
       "dependencies": {
         "@algolia/autocomplete-core": "1.19.2",
+        "@base-ui/react": "^1.5.0",
         "@docsearch/core": "4.6.0",
         "@docsearch/css": "4.6.0",
       },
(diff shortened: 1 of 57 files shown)


LorrisSaintGenez:
Comments are sorted in order of me finding them, not in any specific importance order 😅 
These are mostly minor, great work on this redesign 👏 

- (no strong opinion on this one) Should we be able to un-pin a content? Not delete it, just un-pin it
Screenshot 2026-06-09 at 10 22 34

- The Pin icon feels a bit big in the circle, we should reduce the size of the pin while keeping the circle the same size
Screenshot 2026-06-09 at 10 22 59


- Should we limit the # of AI conversations displayed by default? I had to scroll a bit to find the Pinned conversation 
Screenshot 2026-06-09 at 10 25 32


- Feels like there's a bit much spacing between answers and previous question
Screenshot 2026-06-09 at 10 34 14


- Personal opinion, but not a big fan of the change of behavior were I could just press "Enter" to ask AI, and now I need to click on "Ask AI" 😢 or press `cmd+Enter`, but it's not obvious, I had to read your PR description to know it. Would there be a way to add a hint for this?


- No no-result state when Ask AI is active
<table>
<tr>
 <td>With Ask AI
 <td>Without Ask AI
<tr>
 <td>
Screenshot 2026-06-09 at 10 38 53

 <td>
Screenshot 2026-06-09 at 10 38 40

</table>

- "Cancel" button in dark mode is black, should be white 👀 
Screenshot 2026-06-09 at 10 40 57

- Why did we move the AI-generated answers warning to the bottom?

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>Screenshot 2026-06-09 at 10 47 29
 <td>Screenshot 2026-06-09 at 10 47 21
</table>

- Should we by default parse HTML entities? French seems tough on the preview 🇫🇷 
Screenshot 2026-06-09 at 10 55 12

- "Answering" is not horizontally aligned with the Arrow icon
Screenshot 2026-06-09 at 10 55 25

LorrisSaintGenez:
- I have multiple AI conversations, but even when starting a new one, I get hit with thread depth limit 🤔 Can't reproduce the bug on the `v5` branch. When I restart the playground, it "resets" the counter. It feels like it counts the total number of messages in all visible conversations, not per conversation. Happens also in the sidebar - "nah good" was my first message in the sidebar conversation (and `Chat error` is not dark-moded 🕶️ )

Screenshot 2026-06-10 at 08 55 20
Screenshot 2026-06-10 at 09 02 12


- Could we have the "Answers are generated with AI which can make mistakes." left-aligned, like we have today?

Screenshot 2026-06-10 at 09 03 14

- Stroke-width for the Pin icon is greater than other icons'
Screenshot 2026-06-10 at 08 52 43

LorrisSaintGenez (approved):
:shipit:

PR #2898, opened 2026-06-10
chore: move CSS building to lightning css

8bittitan:
### Why

The `@docsearch/css` build was a hand-rolled `build-css.js`: it ran each source file through PostCSS/cssnano, manually concatenated the results, and duplicated the bundle-banner helper that every other package already imports from `scripts/`. More importantly, it shipped whatever CSS syntax was authored with **no browser-target processing at all** — no downleveling, no prefixing — so the published stylesheet's browser support was implicit and unenforced (`stylelint-no-unsupported-browser-features` only warned).

Moving to [Lightning CSS](https://lightningcss.dev/) replaces all three concerns — minification, `@import` bundling, and target-driven transpilation/prefixing — with a single fast tool, and lets composition live in CSS (`@import`) instead of JS string concatenation. A `browserslist: ["defaults"]` field now makes the support contract explicit and shared between the build and stylelint, so prefixes/fallbacks are emitted automatically and consistently.

The published artifacts are intentionally unchanged in shape (same dist files, names, `.scss` copies, banners), so there is **zero blast radius** for consumers, the `@docsearch/react` style shims, examples, docs, and CDN links.

Secondary cleanups that fell out of the rewrite: the script is now TypeScript run via Bun (matching the other packages' build configs), the duplicated banner helper and several `eslint-disable` headers are gone, `postcss`/`cssnano` are dropped from the root, and the build now exits non-zero on failure (the old script swallowed write errors).

### Benefit

- CSS build times `760ms → 325ms`
- The immediate benefit is that the compiled bundle size of `dist/style.css` went `6324 → 6284 bytes`.
- We can now start writing modern CSS in source, which is safely transpiled at build.

### Notable behavior change

The handwritten `-webkit-backdrop-filter` is now dropped from the output because every browser in `defaults` supports the unprefixed property. This only affects Safari/iOS older than 18.5.

Output size is unaffected: `style.css` is 6.07 KB gzip, still under the existing bundlesize cap.

### Testing scenarios

- `bun run build:pkg:css` and confirm `dist/` contains the same files with valid banners; `.scss` copies match their `.css` counterparts.
- `bun run test:size` — `style.css` stays under the 6.15 kB cap.
- Render the modal, button, and sidepanel in a browser (e.g. `examples/demo-react`, `examples/demo-js`, and the website) and confirm styling is unchanged in light/dark themes.
- Verify a Sass consumer importing `@docsearch/css/dist/style.scss` still resolves.
- E2E smoke: `bunx playwright test --project=chromium` against the website with the rebuilt CSS.


dylantientcheu (approved):
🐐 ⚡

PR #2900, opened 2026-06-15
fix(askai): Fix new conversation causing thread depth errors

8bittitan:
### Why

Starting a "new conversation" in Ask AI previously only cleared the rendered messages (`setMessages([])`) while keeping the same underlying chat session alive. Because the conversation id never rotated, the backend kept appending every "new" conversation onto the same thread, which eventually tripped the thread depth limit (AI-217).

This change ties each conversation to its own `chatId`:
- Starting a new conversation rotates the id, recreating the chat session with an empty thread.
- Restoring a stored conversation reuses its persisted `chatId` (or generates a fresh one if none exists) and seeds the prior messages so history is preserved across session recreation.
- The `chatId` is now persisted on stored conversation hits so restores stay anchored to the correct thread.

### What changed (high level)
- `useAskAi` now owns the `chatId` lifecycle and exposes `startNewConversation` / `restoreConversation`, replacing direct `setMessages([])` calls in the modal and sidepanel.
- `chatId` added to `StoredAskAiState` and threaded through `buildDummyAskAiHit`.

### Testing scenarios
1. Open Ask AI, ask a question, click "New conversation", ask several more questions in a row, confirm no thread depth error appears after repeated new conversations.
2. Have a conversation, close/toggle Ask AI off and back on, start a new question, confirm it starts a clean thread rather than appending to the old one.
3. From conversation history, restore a previous conversation, confirm prior messages render and follow-up questions continue the correct thread.
4. Restore an older stored conversation that predates this change (no persisted `chatId`), confirm it still loads and a fresh id is assigned.
5. Repeat the same flows in both the **modal** and the **sidepanel** (both call sites were updated).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,7 +77,7 @@ bun run pw:run:webkit
 
 ### Imports
 
-Imports must be ordered alphabetically with newlines between groups:
+Imported modules must be ordered alphabetically with newlines between groups:
 
 1. Built-in modules
 2. External dependencies
--- a/packages/docsearch-react/src/DocSearchAskAiModal.tsx
+++ b/packages/docsearch-react/src/DocSearchAskAiModal.tsx
@@ -25,6 +25,7 @@ import type {
   InternalDocSearchHit,
   OnAskAiFeedback,
   StoredAskAiMessage,
+  StoredAskAiState,
   SuggestedQuestionHit,
 } from './types';
 import { type AskAiState } from './types/AskiAi';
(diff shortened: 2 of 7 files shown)


LorrisSaintGenez (changes requested):
It doesn't send the message when pressing "Enter" to ask ai

https://github.com/user-attachments/assets/26e1bbf1-9e1c-486d-b7ac-4f36dc144c91

LorrisSaintGenez (approved):

PR #2908, opened 2026-07-07
fix: increase line height to avoid overlap

### LINKED ISSUE #2883: Long search query results overlap when text is quoted and no results found
## Steps to reproduce

1. Click the Docsearch field.
2. Enter a long string in quotes (`"content"`) that won't return results.
3. See overlapping text in results.

Image

## Expected behavior

The "No results found for "content..." is properly wrapped.

## Environment

- OS: macOS
- Browser: Chrome Version 148.0.7778.168
- DocSearch version: v4
- Docusaurus version: 3.9.2

brc-dd:
fixes #2883

before:

(on Docusaurus' website)

image<br />

after:

image<br />

### DIFF
--- a/packages/docsearch-css/src/modal.css
+++ b/packages/docsearch-css/src/modal.css
@@ -316,7 +316,7 @@
   color: var(--docsearch-text-color);
   vertical-align: middle;
   font-size: 1.1em;
-  line-height: 0.5em;
+  line-height: 1.5em;
   font-weight: 300;
 }
 


8bittitan (approved):

SoonIter:
Hi @dylantientcheu, thanks for merging this fix! It is still reproducible in both Rspress and Docusaurus since the latest stable release predates this PR. 
If a 4.6.4 release is planned, would it be possible to include this change? Thanks!

dylantientcheu:
thanks for the ping, we'll trigger a release  today.

PR #2912, opened 2026-07-09
feat(v5): Add prompt suggestions to keyword search

8bittitan:
### Why

When users type a keyword query, they often don't realize they can ask the AI a
question instead, or they aren't sure how to phrase one. Surfacing relevant,
index-backed prompt suggestions alongside the "Ask AI" action nudges users
toward the Ask AI flow with ready-made questions that match what they're
already searching for. This lowers the effort of formulating a good question and
drives more engagement with the assistant.

To support this, sites can point DocSearch at an Algolia index of prompt
suggestions, and matching prompts are fetched as the user types and shown under
the Ask AI action in the keyword results.

### What changed (high level)

- Adds a `promptSuggestions` prop (`{ indexName, hitsPerPage? }`, defaults to 3
  suggestions) that enables fetching prompts from a dedicated Algolia index.
- Fetches suggestions in parallel with the regular Algolia results so keyword
  search is not slowed down, and fails silently if the suggestion request errors.
- Renders the suggestions under the Ask AI section; selecting one submits it as
  the Ask AI question.
- Refactors `buildAskAiActionSources` to be async and share a single
  `createAskAiHit` helper for both the Ask AI button and the suggestions.
- Adds the `askAiResultsTitle` translation key (defaults to "Ask AI Assistant").
- Accessibility: the results list is now labelled by its heading via
  `aria-labelledby`, the section title uses a proper `h2`, and the sparkles icon
  is marked `aria-hidden`.

### Screenshots

Screenshot 2026-07-09 at 4 33 21 PM

### Testing scenarios

1. **Suggestions render**: Configure `promptSuggestions={{ indexName: '...' }}`,
   open the modal, and type a query. Confirm matching prompt suggestions appear
   under the Ask AI action.
2. **Select a suggestion**: Click a suggestion and confirm it is submitted as the
   Ask AI question and the answer starts streaming.
3. **hitsPerPage**: Set `hitsPerPage` to a custom value and confirm the number of
   suggestions shown matches (default is 3 when omitted).
4. **Disabled by default**: Omit `promptSuggestions` entirely and confirm only the
   Ask AI action shows, with no extra index request fired.
5. **Graceful failure**: Point `indexName` at a non-existent/erroring index and
   confirm the Ask AI action still renders without breaking keyword search.
6. **i18n**: Override `askAiResultsTitle` and confirm the custom section title
   renders.
7. **Accessibility**: Confirm the suggestions list is announced with the section
   heading (via `aria-labelledby`) and that the sparkles icon is hidden from
   screen readers.

### DIFF
--- a/packages/docsearch-react/src/utils/createAskAiSources.ts
+++ b/packages/docsearch-react/src/utils/createAskAiSources.ts
@@ -10,22 +10,27 @@ import { SOURCE_IDS } from './collections';
 
 const MAX_RECENT_CONVERSATIONS_DISPLAYED = 3;
 
+const EMPTY_RESULT_CONTENT = {
+  value: '',
+  matchLevel: 'none',
+  matchedWords: [],
+} satisfies InternalDocSearchHit['_highlightResult']['content'];
 const EMPTY_HIERARCHY_HIGHLIGHT_RESULT = {
-  lvl0: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl1: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl2: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl3: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl4: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl5: { value: '', matchLevel: 'none', matchedWords: [] },
-  lvl6: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl0: EMPTY_RESULT_CONTENT,
+  lvl1: EMPTY_RESULT_CONTENT,
+  lvl2: EMPTY_RESULT_CONTENT,
+  lvl3: EMPTY_RESULT_CONTENT,
+  lvl4: EMPTY_RESULT_CONTENT,
+  lvl5: EMPTY_RESULT_CONTENT,
+  lvl6: EMPTY_RESULT_CONTENT,
 } satisfies InternalDocSearchHit['_highlightResult']['hierarchy'];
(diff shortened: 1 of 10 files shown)


packages/docsearch-react/src/utils/createAskAiSources.ts
```diff
@@ -1,12 +1,35 @@
...
+  lvl0: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl1: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl2: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl3: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl4: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl5: { value: '', matchLevel: 'none', matchedWords: [] },
+  lvl6: { value: '', matchLevel: 'none', matchedWords: [] },
+} satisfies InternalDocSearchHit['_highlightResult']['hierarchy'];
+const EMPTY_HIGHLIGHT_RESULT: InternalDocSearchHit['_highlightResult'] = {
+  content: { value: '', matchLevel: 'none', matchedWords: [] },
```
LorrisSaintGenez: Nit: Could we factorize this `{ value: '', matchLevel: 'none', matchedWords: [] }`? 👀

LorrisSaintGenez (approved):

packages/docsearch-react/src/utils/createAskAiSources.ts
8bittitan: Done!

dylantientcheu (approved):

PR #2913, opened 2026-07-09
chore(v5): Split Ask AI related CSS into own bundle

8bittitan:
## Summary

The Ask AI feature ships a large amount of CSS that every DocSearch user downloads, even when they don't enable Ask AI. This splits the Ask AI styles out of the main `modal.css`/`style.css` bundle into a dedicated `askai.css` bundle so consumers only pay for those styles when they actually use the feature.

This drops the core `style.css` budget from 8 kB down to 5.5 kB, with the extracted Ask AI styles landing in a separate 4.5 kB bundle.

## What changed

- Moved all Ask AI related rules out of `modal.css` and into a new `src/askai.css` source file
- Added an `askai` entry to the CSS build so it emits `dist/askai.css` and `dist/askai.scss`
- Exposed new granular style entry points from `@docsearch/react` (`./style/askai`, `./style/button`, `./style/modal`, `./style/variables`) so consumers can opt into individual style bundles
- Updated bundlesize budgets to track the two bundles separately

## Impact

- Default `style.css` size reduced to `5.04 kB` from `7.9 kB`

## Testing

- Run the React playground (`bun run playground:start`) and open the Ask AI example. With `@docsearch/css/dist/askai.css` imported, the Ask AI panel should render fully styled.
- Temporarily remove the `askai.css` import in `examples/demo-react/src/App.tsx` and confirm the base modal still looks correct while the Ask AI panel loses its styling, verifying the split is clean.
- Run `bun run test:size` to confirm the new `style.css` (5.5 kB) and `askai.css` (4.5 kB) budgets pass.
- Import `@docsearch/react/style/askai` in a consumer app and confirm the styles resolve correctly through the new export path.


LorrisSaintGenez (changes requested):
A couple of things 👀 
- Sidepanel is now missing a few classes that were moved to `askai.css` (mostly user message and agent states)
- We should probably update the docs to mention the new `askai.css`

(Ignore the error)

<table>
<tr>
 <td>Live Sidepanel
 <td>This PR' Sidepanel
<tr>
 <td>
Image
 <td>
Image
</table>

dylantientcheu (commented):
Absolutely the same as Lorris said.

in `packages/docsearch-css/src/style.css`

Default CSS entry omits askai.css, breaking Ask AI styling for existing consumers.

8bittitan:
@dylantientcheu "Default CSS entry omits askai.css, breaking Ask AI styling for existing consumers." - I think this is fine, `v5` is a breaking change, and IMO the defaults shouldn't be the bloated bundle. Def open to other opinions though.

8bittitan:
> A couple of things 👀
> 
> * Sidepanel is now missing a few classes that were moved to `askai.css` (mostly user message and agent states)
> * We should probably update the docs to mention the new `askai.css`
> 
> (Ignore the error)
> 
> Live Sidepanel	This PR' Sidepanel
> Image	Image

@LorrisSaintGenez 

- The Sidepanel one seems like it's more from the adapter being behind, it should be fixed with those updates
- There's a larger documentation ticket later to handle docs

8bittitan:
> Absolutely the same as Lorris said.
> 
> in `packages/docsearch-css/src/style.css`
> 
> Default CSS entry omits askai.css, breaking Ask AI styling for existing consumers.

@dylantientcheu Decided to go this route. `style.css` now bundles modal+askai styling, `sidepanel.css` is still standalone. `button.css` and `modal.css` are still there for those that want the slimmer bundle size.

LorrisSaintGenez (approved):
Looking good!

PR #2915, opened 2026-07-10
fix: Ensure stage level and watch level scripts use bun runtime

8bittitan:
`bun run build` was failing after recently moving to use `nodejs latest`.

This change just ensures that specific scripts are using the bun runtime so that tsdown configs can properly import helper files.


package.json
```diff
@@ -17,9 +17,9 @@
     "build:pkg:sidepanel-js": "bun --filter @docsearch/sidepanel-js build",
     "build:pkg:modal": "bun --filter @docsearch/modal build",
     "build:pkg:adapter": "bun --filter @docsearch/docusaurus-adapter build",
-    "build:stage:base": "bun run --parallel build:pkg:core build:pkg:css",
-    "build:stage:react": "bun run build:pkg:react",
-    "build:stage:consumers": "bun run --parallel build:pkg:js build:pkg:sidepanel build:pkg:sidepanel-js build:pkg:modal",
+    "build:stage:base": "bun run -b --parallel build:pkg:core build:pkg:css",
+    "build:stage:react": "bun run -b build:pkg:react",
+    "build:stage:consumers": "bun run -b --parallel build:pkg:js build:pkg:sidepanel build:pkg:sidepanel-js build:pkg:modal",
     "build:stage:adapter": "bun run build:pkg:adapter",
```
LorrisSaintGenez, line 24: This one doesn't need the `-b` flag?

LorrisSaintGenez (commented):
@dylantientcheu I know we just updated to `nodejs latest` but couldn't we go back to 22.15.0 or there's a specific reason for `latest`? 👀  We wouldn't need to do this PR and also we'd keep the same version as the one used in our circleci config

package.json
8bittitan: It is not, the adapter(s) don't use `tsdown` to bundle, good eye though 👍

LorrisSaintGenez (approved):

PR #2916, opened 2026-07-10
feat(js): Document JS based hybrid mode, fix JS packages

8bittitan:
## Summary

Hybrid Mode (Sidepanel running alongside the DocSearch Modal) previously only worked with the React integration, and the docs explicitly called out that the vanilla JS path was unsupported. This PR makes Hybrid Mode actually usable from the JS-only packages and documents how to wire it up.

The JS packages were not bundling their internal dependencies (`@docsearch/react` and `@docsearch/core`), which left them broken for consumers who don't already have those packages resolved. Marking them as `noExternal` in the tsdown config bundles them in so the JS entry points work standalone.

The demo was also updated so the Modal and Sidepanel close each other when one opens, which is the expected Hybrid Mode behavior and doubles as the reference implementation shown in the docs.

## Why

- The JS builds shipped with unresolved internal imports, so `@docsearch/js` and `@docsearch/sidepanel-js` did not work on their own.
- Users had no guidance for setting up Hybrid Mode outside of React, and the docs claimed it wasn't possible at all.

## Testing

- Build `@docsearch/js` and `@docsearch/sidepanel-js` and confirm the output bundles include the core/react code rather than leaving them as external imports.
- Run the JS demo (`bun run playground-js:start`) and:
  - Trigger Ask AI from the Modal and confirm the Modal closes and the Sidepanel opens with the conversation carried over.
  - Open the Sidepanel and confirm the Modal closes if it was open, and vice versa.
- Follow the new JavaScript tab in the Hybrid Mode docs in a fresh project and confirm the setup works end to end.

### DIFF
--- a/packages/website/docs/sidepanel/hybrid.mdx
+++ b/packages/website/docs/sidepanel/hybrid.mdx
@@ -152,6 +152,9 @@ docsearchInstance = docsearch({
     docsearchInstance?.close();
     sidepanelInstance.open(initialMessage);
     return true;
+  },
+  onOpen: () => {
+    sidepanelInstance?.close();
   }
 });
 
--- a/packages/website/docs/sidepanel/hybrid.mdx
+++ b/packages/website/docs/sidepanel/hybrid.mdx
@@ -5,10 +5,6 @@ title: Hybrid Mode
 import Tabs from  '@theme/Tabs';
 import TabItem from '@theme/TabItem';
 
-:::info
-Currently Hybrid Mode is only available when using the React usage approach. Hybrid Mode is not available in the JavaScript-only (vanilla) integration.
-:::
-
 ## Introduction
 
 Sidepanel can run alongside the DocSearch Modal through what we call "Hybrid Mode." When a user initiates an Ask AI action from within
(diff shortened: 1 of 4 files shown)


packages/website/docs/sidepanel/hybrid.mdx
```diff
@@ -97,4 +112,53 @@ function HybridMode() {
...
+  indexName: "YOUR_INDEX_NAME",
+  appId: "YOUR_APP_ID",
+  apiKey: "YOUR_SEARCH_API_KEY",
+  askAi: "YOUR_ASSISTANT_ID",  // Or configuration object
+  interceptAskAiEvent: (initialMessage) => {
+    docsearchInstance?.close();
+    sidepanelInstance.open(initialMessage);
+    return true;
+  }
+});
```
LorrisSaintGenez, line 159: Do we need the `onOpen` here too to close the sidepanel?

LorrisSaintGenez (approved):
Just one question, all good otherwise

brc-dd:
`noExternal` is now deprecated, you should use `deps.alwaysBundle` instead 👀 (https://tsdown.dev/options/dependencies#migration-from-deprecated-options)

packages/website/docs/sidepanel/hybrid.mdx
8bittitan: Ah good catch, added!

8bittitan:
> `noExternal` is now deprecated, you should use `deps.alwaysBundle` instead 👀 (https://tsdown.dev/options/dependencies#migration-from-deprecated-options)

@brc-dd Thanks for pointing that out! I'll do that as a follow-up PR bumping some of our deps versions 😄

PR #2904, opened 2026-07-02
refactor(docusaurus-adapter): rework theme config for v5 and modularize SearchPage

dylantientcheu:
## Summary
- Drops legacy `themeConfig.algolia` support in favor of `themeConfig.docsearch` only, replacing `searchPagePath` with `searchPage.path` / `searchPage: false`
- Removes the `lodash`/`utility-types` dependencies from the adapter and adds `@docsearch/core`, `@docsearch/modal`, `@docsearch/sidepanel`, and `search-insights` as direct deps
- Splits the monolithic `SearchPage` into dedicated `components/`, `hooks/`, `constants.ts`, `types.ts`, and `utils.ts` for maintainability
- Tightens `validateThemeConfig` schemas (e.g. `.unknown(false)` on side panel schemas) and updates docs/examples accordingly
- Adds `adapters/**/src/**/*.test.ts` to the Vitest include globs so adapter tests run in the workspace test suite

## Test plan
- [x] `bun run test --run adapters/docusaurus-theme-search-algolia/src/__tests__/validateThemeConfig.test.ts` passes (23/23)
- [x] `bun run test:types`
- [x] `bun run lint`
- [x] Manual smoke test of the Docusaurus search page (search, facets, Ask AI side panel) against the updated theme config

PR #2917, opened 2026-07-14
feat(website): one-command CLI install + Segment install-intent events on MCP docs

dylantientcheu:
## What

Reworks the **Install DocSearch MCP** docs (`docs/mcp/installation`) so we can see who actually wants to install it, and gives the fastest path.

### One-command install (design)

- New terminal-style **One-command install** section in `MCPInstall`, between the client picker and the step-by-step panel — `npx @docsearch/cli setup --<client>`, with the flag crossfading as you switch clients (mirrors the MCP frontend hero).
- Renders clean for clients the CLI actually supports (`cursor`, `claude`, `codex`, `opencode` — verified against `packages/docsearch-cli` flags) and shows default for the rest.
- `installation.mdx` now introduces the CLI as the fastest path, with plugin/manual as the do-it-yourself routes.

### Segment install-intent events

- Adds `@segment/analytics-next` behind a small lazy-loaded, SSR-safe helper (`src/lib/segment.js`) — Segment only loads on the first tracked interaction.
- Events (same source as the MCP frontend, so both sites funnel together; `context.page` tells them apart):
  - `Install Command Copied` — CLI one-liner copy (`client`, `command`)
  - `Install Config Copied` — any config code-card copy (`client`, `mode`, `caption`)
  - `Install Button Clicked` — Cursor one-click deeplink (`client`, `mode`)

## Why based on `feat/docusaurus-adapter-v5`

The website only builds with the migrated `themeConfig` shape from #2904, so this stacks on top of it and should merge after it.

## Test plan

- `bun run build` in `packages/website` — client + SSR compile, static build succeeds
- Not yet manually exercised in a browser — worth a quick pass over the picker (flag follows the selected client, section collapses for unsupported clients) and the Segment debugger before merging

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/website/src/components/mcp/MCPInstall.jsx
```diff
@@ -577,6 +647,20 @@ export default function MCPInstall() {
           })}
         </motion.div>
 
+        {/* Fastest path: supported clients include their setup flag; others use the interactive setup. */}
+        <AnimatePresence initial={false}>
```
vascobettencourt, line 651: This `AnimatePresence` always receives a child, so the quick-install panel never exits for unsupported clients. Selecting ChatGPT, Pi, VS Code, Windsurf, Zed, Conductor, Antigravity, or Other still shows `npx @docsearch/cli setup` and claims it adds the server/rules/skills to that selected client. The unflagged CLI only detects/prompts for Cursor, Claude, Codex, and OpenCode, so it cannot configure the selected unsupported client. Please gate the child on `CLI_SUPPORTED_CLIENTS.has(selected.id)` (which will also let this exit animation run), as described in the PR summary.

vascobettencourt (changes requested):
The quick-install path needs to be limited to the clients the CLI can actually configure. CI is green, but the current UI advertises a nonfunctional install command for every unsupported client.

packages/website/src/components/mcp/MCPInstall.jsx
dylantientcheu: by default the `setup` command installs for most agents in the relevant folders.

those that are not listed is because it will justwork™️  without any specific install instructions

vascobettencourt (approved):

PR #2918, opened 2026-07-14
chore(tsdown): Bump to latest tsdown version

8bittitan:
## Why

Our tsdown setup was pinned to an older release and relied on the standalone `replacePlugin` from rolldown for build-time defines. The latest tsdown ships native `define` support and a first-class `exports` config, so this upgrade lets us drop the extra plugin wiring and move all packages onto the current toolchain.

Bumping tsdown also required renaming the config files to `.mts` and updating the CI/Cursor Cloud toolchain to Node 24.13.1 to match `.tool-versions`.

## What changed

- Bumped tsdown to the latest version across all packages
- Replaced the rolldown `replacePlugin` in `tsdown.base.ts` with native `define` and a shared `pkgExports` config
- Migrated per-package `tsdown.config.ts` files to `.mts`
- Moved CI (`.circleci/config.yml`) and the Cursor Cloud notes in `AGENTS.md` to Node 24.13.1
- Re-baselined `bundlesize.config.json` limits to reflect the new bundle output, and added entries for the sidepanel packages

## Testing

- Build the libraries and confirm they complete without the removed plugin:
  ```
  bun run --sequential build:stage:base build:stage:react build:stage:consumers build:stage:adapter
  ```
- Run `bun run test --run` and `bun run test:types` to confirm nothing regressed
- Run `bun run test:size` to verify every package stays under the updated bundlesize limits
- Start the playground (`bun run playground:start`) and the JS demo (`bun run playground-js:start`) and confirm search, Ask AI, and the sidepanel still render and work against live results
- Confirm the ESM/UMD output still exposes the expected entry points (`index`, `button`, `modal`, `sidepanel`, `docsearchAi`, `askaiModal`, styles)


adapters/docusaurus-theme-search-algolia/package.json
```diff
@@ -52,10 +52,10 @@
     "utility-types": "^3.10.0"
   },
   "devDependencies": {
-    "@docusaurus/core": "^3.9.2",
-    "@docusaurus/module-type-aliases": "3.9.2",
-    "@docusaurus/theme-classic": "3.9.2",
-    "@docusaurus/types": "^3.9.2",
+    "@docusaurus/core": "^3.10.2",
```
dylantientcheu, line 55: Careful with these updates

If we want to support 3.9 and upwards we should keep it ^3.9.

Is there any reason why we need to bump this?

examples/demo-react/package.json
```diff
@@ -11,6 +11,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
+    "@algolia/autocomplete-core": "1.19.2",
```
dylantientcheu, line 14: Do we need this as a dependency everytime now ?

dylantientcheu (commented):
A few questions

examples/demo-react/package.json
8bittitan: We only need it here because we directly import a type from the package in `demo-react` it's not needed for normal installs.

adapters/docusaurus-theme-search-algolia/package.json
8bittitan: I started getting a weird schema error during build.

But also with `v5`, maybe it's fine to drop `3.9` support since it's stated to not be actively maintained 🤔

dylantientcheu: ok let's drop it

dylantientcheu (approved):

PR #2919, opened 2026-07-15
feat(askai): Move askai related props under root askai

8bittitan:
## Summary

The Ask AI configuration surface had drifted apart. Props like `tools`, `memory`, and `promptSuggestions` lived at the root of `DocSearchAIProps`, while the rest of the Ask AI settings (`assistantId`, `suggestedQuestions`, `indices`, `searchParameters`) were nested under `askAi`. This split forced consumers to reason about two different places to configure a single feature, and made the Docusaurus adapter harder to keep in sync.

This change consolidates all Ask AI related options under the root `askAi` object so there is one clear, predictable place to configure the assistant. The Docusaurus adapter, validation schema, examples, and website config are updated to match.

## Why

- A single `askAi` object is easier to discover and document than options scattered across the props root.
- Keeping the adapter's `themeConfig.docsearch.askAi` shape aligned with the React API reduces the mapping logic and the chance of drift.
- Validation now covers `tools`, `memory`, and `promptSuggestions` so misconfigurations surface at build time in Docusaurus rather than at runtime.

## Breaking change

Consumers passing `tools`, `memory`, or `promptSuggestions` directly on `DocSearchAI` need to move them inside `askAi`:

```tsx
// Before
<DocSearchAI
  askAi={{ assistantId: '...' }}
  tools={customTools}
  promptSuggestions={{ indexName: '...' }}
/>

// After
<DocSearchAI
  askAi={{
    assistantId: '...',
    tools: customTools,
    promptSuggestions: { indexName: '...' },
  }}
/>
```

## Testing scenarios

- In the React playground (`bun run playground:start`), open the Ask AI demo and confirm custom `tools` still render and fire `onToolCall` (the console message tool should log and echo back).
- Trigger a keyword search and verify prompt suggestions still appear now that `promptSuggestions` is nested under `askAi`.
- Enable `memory` under `askAi` with a `userToken` and confirm memory driven tool calls behave as before.
- In a Docusaurus site, set `themeConfig.docsearch.askAi.tools`, `.memory`, and `.promptSuggestions` and confirm the config passes validation and forwards through to the side panel.
- Pass an unknown key under `askAi` and confirm validation rejects it.
- Run `bun run test --run` and `bun run test:types` to confirm the API and validation tests pass.

### DIFF
--- a/adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
+++ b/adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
@@ -356,8 +356,10 @@ function DocSearch({
             appId={sidePanelAskAi.appId}
             indexName={sidePanelAskAi.indexName}
             searchParameters={sidePanelAskAi.searchParameters}
-            indices={sidePanelAskAi.indices}
+            indices={panelOptions.indices ?? sidePanelAskAi.indices}
             suggestedQuestions={panelOptions.suggestedQuestions ?? sidePanelAskAi.suggestedQuestions}
+            tools={panelOptions.tools ?? sidePanelAskAi.tools}
+            memory={panelOptions.memory ?? sidePanelAskAi.memory}
           />
         )}
     </>
--- a/adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
+++ b/adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
@@ -20,6 +20,7 @@ import type {
   DocSearchTranslations,
   InternalDocSearchHit,
   StoredDocSearchHit,
+  ToolCalls,
 } from '@docsearch/react';
 import { SidepanelButton } from '@docsearch/sidepanel/button';
 import type { Sidepanel as SidepanelType } from '@docsearch/sidepanel/sidepanel';
@@ -43,12 +44,13 @@ import {
   useSearchResultUrlProcessor,
 } from '../../client';
 
-import type { ThemeConfigDocSearch } from '@docsearch/docusaurus-adapter';
+import type { DocusaurusSearchBarAskAiProps, ThemeConfigDocSearch } from '@docsearch/docusaurus-adapter';
 
 type NavigatorNavigateParams = Parameters<NonNullable<NonNullable<DocSearchModalProps['navigator']>['navigate']>>[0];
 
-type SidePanelOptions = Exclude<NonNullable<ThemeConfigDocSearch['sidePanel']>, boolean>;
+type SidePanelOptions = Exclude<NonNullable<ThemeConfigDocSearch['sidePanel']>, boolean> & { tools?: ToolCalls };
 type SidePanelPanelOptions = Omit<SidePanelOptions, 'hideButton' | 'keyboardShortcuts'>;
+type AskAiOptions = NonNullable<ThemeConfigDocSearch['askAi']> & Pick<DocusaurusSearchBarAskAiProps, 'tools'>;
 
 type AdapterDocSearchProps = Omit<
   DocSearchAskAiModalProps,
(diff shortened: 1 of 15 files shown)


adapters/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
```diff
@@ -112,11 +81,65 @@ const AskAiIndexSchema = Joi.object({
   searchControls: AgentStudioSearchControlsSchema.optional(),
 }).unknown(false);
 
+const AskAiToolDefinitionSchema = Joi.object({
```
LorrisSaintGenez: **Blocking — `tools` via themeConfig can't survive a production build.**

Docusaurus emits the client config with `JSON.stringify(siteConfig)` (`@docusaurus/core/lib/server/codegen/codegen.js`, `genSiteConfig`), which silently drops function-valued properties. Joi validation runs server-side on the raw config *before* codegen, so `render`/`onToolCall` pass validation, then reach the client as `{ myTool: {} }` — a truthy object.

Runtime result, with zero build-time error:
- render-only tool → `tool.render(...)` throws in `ToolCall.tsx` (`CustomTool`, no guard on `render`)
- tool with `onToolCall` → `useAskAi.ts` `handleToolCall` returns silently → `addToolOutput` never fires → permanent "Loading tool" shimmer, conversation stalls

Suggestion: drop `tools` from `AskAiSchema`/`SidePanelSchema` and the `.d.ts` (keep `memory`/`promptSuggestions` — plain JSON, they're fine), and document that `tools` requires swizzling `SearchBar`. Or keep the key but reject it with an explicit Joi error explaining why. Keeping the `tools={panelOptions.tools ?? sidePanelAskAi.tools}` forwarding in `SearchBar/index.tsx` is right — it works for the swizzle path.

Question: was the Docusaurus test scenario run with `docusaurus build`, or dev only?

adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
```diff
@@ -356,8 +356,10 @@ function DocSearch({
             appId={sidePanelAskAi.appId}
             indexName={sidePanelAskAi.indexName}
             searchParameters={sidePanelAskAi.searchParameters}
-            indices={sidePanelAskAi.indices}
+            indices={panelOptions.indices ?? sidePanelAskAi.indices}
```
LorrisSaintGenez, line 361: `panelOptions.indices` is unreachable through themeConfig: `SidePanelSchema` is `.unknown(false)` and doesn't declare `indices`, so Joi rejects `sidePanel.indices` — while the TS type (via the new `Pick` in `SidepanelProps`) says it's legal. Either add `indices` to `SidePanelSchema` or drop this fallback.

e2e/search.spec.ts
```diff
@@ -91,6 +91,9 @@ test.describe('Search', () => {
     await expect(docSearch.hits).toBeVisible();
     await page.keyboard.press('ArrowDown');
     await page.keyboard.press('ArrowDown');
+    await page.keyboard.press('ArrowDown');
```
LorrisSaintGenez: The 5-down/1-up count hardcodes "1 Ask AI button + exactly 3 prompt suggestions" returned by the live `docsearch-markdown_prompt_suggestions` index for query `g`. If that index changes, this flakes or selects a suggestion instead of a hit. Prefer targeting the first real hit with a locator.

LorrisSaintGenez (commented):
Reviewed the askAi consolidation — the prop move itself is clean and complete (no leftover consumers of the removed root props across the monorepo). One blocking concern on `tools` via themeConfig, two smaller points inline.

LorrisSaintGenez (changes requested):
Marking as request-changes per the blocking comment above (tools via themeConfig stripped by JSON.stringify at build).

e2e/search.spec.ts
8bittitan: Targeting a real hit through a selector defeats the point of the test though 😅 which is keyboard navigation.

8bittitan: Nvm, I got `selector` and `locator` mixed up, fixed 👍

adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx
8bittitan: Done 👍

adapters/docusaurus-theme-search-algolia/src/validateThemeConfig.ts
8bittitan: Dropped from the public config schema, documented how to use it with Swizzle.

LorrisSaintGenez (approved):

dylantientcheu (approved):
approving, but on a siiiiiide note, i think its okay for default docusaurus to not support client tools and we can let users swizzle the search bar if they really need it.

PR #2927, opened 2026-07-17
fix: Agent Studio MCP search tool

8bittitan:
## Summary

The Agent Studio MCP search tool changed its input shape. It now sends a `queries` array (with `originalQuery` and `clickAnalytics`) instead of a single top-level `query` string. Because the UI only knew how to read the old single-query input, MCP search calls rendered blank query labels and multi-query searches were collapsed into a single (often empty) entry.

This updates the type definitions and rendering to support the new multi-query payload while still parsing older stored tool calls that use the legacy single-query shape, so previously saved conversations keep rendering correctly.

## Why

- The MCP tool now issues multiple queries per call, and each one should be shown and be individually clickable.
- Older conversations persisted with the previous input shape must not break, so query extraction handles both formats.
- Centralizing query extraction in a single `getSearchToolQueries` helper removes the duplicated, format-specific parsing that was scattered across the tool call rendering and grouping logic.

## Testing

- Ask AI a question that triggers the Agent Studio MCP search tool and confirm each issued query renders as its own result row.
- Confirm clicking (and keyboard Enter/Space on) a rendered query still fires the search-for-query action.
- Trigger a call that returns multiple queries and verify empty or whitespace-only queries are filtered out.
- Open an older conversation that used the previous single-query MCP payload and confirm the query still displays.
- Run `bun run test --run packages/docsearch-react/src/components/__tests__/ToolCall.test.tsx` and the `ai` / `groupConsecutiveToolResults` test files.

### DIFF
--- a/.changeset/sharp-donkeys-search.md
+++ b/.changeset/sharp-donkeys-search.md
@@ -1,0 +1,20 @@
+---
+"@docsearch/react": patch
+---
+
+fix(askai): support Agent Studio's batched MCP search tool
+
+The Algolia MCP search tool can now issue multiple queries in a single tool
+call (`queries[]`), which previously rendered as an empty query and broke
+consecutive tool-call aggregation.
+
+- `AlgoliaMCPSearchTool["input"]` accepts both the legacy single-query shape
+  and the new batched `queries[]` shape, and tolerates a missing `output`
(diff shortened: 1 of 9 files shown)


dylantientcheu (approved):
funny that the test diffs are even more than the actual code :D

PR #2928, opened 2026-07-20
[docsearch-react] Bug Fix: Preserve searchSuggestions array shape for no-results suggestions

### LINKED ISSUE #2922: "Try searching for" no-results suggestions never render in v4 (searchSuggestions built as an object, consumed as an array)
## Description

The "Try searching for" suggestions on the no-results screen never render in v4 (they did in v3). `state.context.searchSuggestions` is written as an object but consumed as an array, so `NoResultsScreen`'s `searchSuggestions.length > 0` guard is always false and the list is silently skipped.

Producer -- `DocSearchModal.tsx` (4.6.3 and `main`):

```ts
searchSuggestions: { ...(sourcesState.context.searchSuggestions ?? []), ...Object.keys(sources) }
// object-literal spread -> { 0: "...", 1: "..." }, not an array
```

Consumer -- `NoResultsScreen.tsx` (4.6.3 and `main`): typed `string[]`, uses `.length` and `.slice`. v3 stored a plain array (`searchSuggestions: Object.keys(sources)`), which is why it worked there.

Fix: array spread instead of object spread (keeps the accumulate-across-searches intent):

```ts
searchSuggestions: [ ...(sourcesState.context.searchSuggestions ?? []), ...Object.keys(sources) ]
```

## Steps to reproduce

1. Open the DocSearch modal.
2. Type a query that returns results (this populates the suggestions with the result group `lvl0`s).
3. Type a query that returns no results.
4. The "Try searching for:" list is missing (it appears in v3).

## Expected behavior

The no-results screen shows the "Try searching for:" suggestions, as it does in v3.

## Environment

- OS: any (code-path bug, not environment specific)
- Browser: any
- DocSearch version: 4.6.3 (same code on `main`)

Fenz-art:
## Description

Fixes an issue where "Try searching for" suggestions were not displayed on the no-results screen.

The issue was caused by `searchSuggestions` being converted from an array into an object-like structure when updating the state in `DocSearchModal`.

`NoResultsScreen` expects `searchSuggestions` to remain an array and uses array operations such as `.length` and `.slice()`. Because the value was stored with the wrong shape, the no-results suggestions were silently skipped.

Closes #2922

## Changes

- Updated `DocSearchModal` to preserve `searchSuggestions` as an array by using array spread instead of object spread.
- Added a regression test covering the previous successful search → no-results flow.

## Root Cause

Before:

```ts
searchSuggestions: {
  ...(previousSuggestions ?? []),
  ...Object.keys(sources)
}


This converted the array into an object:

{
  0: "Suggestion A",
  1: "Suggestion B"
}

After:

searchSuggestions: [
  ...(previousSuggestions ?? []),
  ...Object.keys(sources)
]

This keeps the expected array structure:

[
  "Suggestion A",
  "Suggestion B"
]
Test Plan
Before
Open the DocSearch modal.
Perform a search that returns results to populate suggestions.
Search for a query that returns no results.
Verify that the "Try searching for" suggestions are missing.
After

Added a regression test verifying that previous search suggestions are displayed on the no-results screen.

Command:

yarn vitest run packages/docsearch-react/src/__tests__/api.test.tsx -t "renders no-results suggestions from previous successful searches"

Result:

✓ 1 passed
✓ 14 skipped

The targeted regression test passes with this change.

Note: The full API test file still reports an unrelated Ask AI failure in this environment.


---

## Commit message

```bash
fix(docsearch-react): preserve searchSuggestions array shape

### DIFF
--- a/packages/docsearch-react/src/DocSearchModal.tsx
+++ b/packages/docsearch-react/src/DocSearchModal.tsx
@@ -201,12 +201,13 @@ const buildQuerySources = async ({
 
       // We store the `lvl0`s to display them as search suggestions
       // in the "no results" screen.
-      if ((sourcesState.context.searchSuggestions as any[]).length < Object.keys(sources).length) {
+      const existingSearchSuggestions = Array.isArray(sourcesState.context.searchSuggestions)
+        ? sourcesState.context.searchSuggestions
+        : [];
+
+      if (existingSearchSuggestions.length < Object.keys(sources).length) {
         setContext({
-          searchSuggestions: {
-            ...(sourcesState.context.searchSuggestions ?? []),
-            ...Object.keys(sources),
-          },
+          searchSuggestions: [...existingSearchSuggestions, ...Object.keys(sources)],
         });
       }
 
--- a/packages/docsearch-react/src/__tests__/api.test.tsx
+++ b/packages/docsearch-react/src/__tests__/api.test.tsx
@@ -1,5 +1,5 @@
 import { render, act, fireEvent, screen, cleanup } from '@testing-library/react';
-import type { LiteClient, SearchResponse, SearchResponses } from 'algoliasearch/lite';
+import type { Hit, LiteClient, SearchResponse, SearchResponses } from 'algoliasearch/lite';
 import React, { type JSX } from 'react';
 import { describe, it, expect, afterEach } from 'vitest';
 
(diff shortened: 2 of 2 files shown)


8bittitan:
Hi @Fenz-art we recently released `5.0.0` and in it the building of the sources moved. If you have the time to resolve the conflicts I'd be happy to merge!

PR #2952, opened 2026-08-04
feat(v5): Add new footerAction prop

8bittitan:
## Summary

DocSearch previously had no supported way to place a custom action in the modal footer. Integrators who wanted to surface things like a feedback link, a link to a related project, or a support shortcut had to fork the footer or hack around the DOM. This adds a first-class `footerAction` prop across both `@docsearch/react` and `@docsearch/js` so that content can be rendered inline with the "Powered by Algolia" logo without touching internals.

While wiring this up, the template typings between `@docsearch/react` and `@docsearch/js` had drifted apart, which made the JS wrapper harder to type correctly and led to duplicated inline type definitions. This aligns those types by extracting shared `HitComponentProps` and `ResultsFooterComponentProps` interfaces and centralizing the JS template function types (`HitComponentFn`, `ResultsFooterComponentFn`, `FooterActionFn`), so all three template hooks follow the same shape.

The footer markup was also restructured. It now wraps the logo and the new action in a `.DocSearch-Footer-Actions` container so the two can sit side by side, replacing the old `row-reverse` layout that assumed the logo was the only trailing element.

## What changed

- New `footerAction` prop in `@docsearch/react`, rendered inside `.DocSearch-Footer-Action` before the Algolia logo, in both the standard and Ask AI modals.
- `@docsearch/js` support for `footerAction` using the existing template patterns (html helper, JSX, or function based).
- Shared and consolidated template types between the React and JS packages.
- Footer CSS reworked around a `.DocSearch-Footer-Actions` wrapper, with the logo pushed to the end when no action is present.
- Updated React and JS API reference docs.

## Testing scenarios

- Open the modal with a `footerAction` set and confirm the custom element renders to the left of the Algolia logo.
- Set a `footerAction` that returns `null` and confirm nothing extra renders and the logo stays aligned to the end of the footer.
- Trigger the Ask AI modal with a `footerAction` set and confirm it also renders there.
- In `@docsearch/js`, pass `footerAction` returning a string, a JSX/Preact element, and a component function, and confirm each renders correctly.
- Resize below and above 768px to confirm the footer actions layout switches between full width spacing and inline alignment.
- Render without a `footerAction` and confirm the footer matches previous behavior.

### DIFF
--- a/packages/docsearch-js/src/__tests__/createDocSearch.test.tsx
+++ b/packages/docsearch-js/src/__tests__/createDocSearch.test.tsx
@@ -1,21 +1,16 @@
 import { createElement, type JSX } from 'preact';
 import { afterEach, describe, expect, it } from 'vitest';
 
-import { createDocSearch, type DocSearchInstance } from './createDocSearch';
-import type { FooterActionFn as KeywordFooterActionFn } from './docsearch';
-
-import type { FooterActionFn as AskAiFooterActionFn } from './index';
+import { createDocSearch, type DocSearchInstance } from '../createDocSearch';
+import type { FooterActionFn as KeywordFooterActionFn } from '../docsearch';
+import type { FooterActionFn as AskAiFooterActionFn } from '../index';
 
 interface TestComponentProps {
-  footerAction?: () => JSX.Element | null;
+  footerAction?: JSX.Element | null;
 }
 
 function TestComponent({ footerAction }: TestComponentProps): JSX.Element {
-  return createElement(
-    'div',
-    null,
-    footerAction && createElement(footerAction, null)
-  );
+  return createElement('div', null, footerAction);
 }
 
 const askAiFooterAction: AskAiFooterActionFn = (_props, { html }) =>
--- a/packages/docsearch-js/src/createDocSearch.tsx
+++ b/packages/docsearch-js/src/createDocSearch.tsx
@@ -29,7 +29,7 @@ export interface DocSearchCallbacks {
   interceptAskAiEvent?: (initialMessage: InitialAskAiMessage) => boolean | void;
 }
 
-export type TemplateHelpers = Record<string, unknown> & { html: typeof html };
+export type TemplateHelpers = { html: typeof html };
 
 // Defines the public facing interface for each "template" function
 type TemplateFnReturnType = JSX.Element | string | (() => JSX.Element) | null;
@@ -45,7 +45,7 @@ export type ResultsFooterComponentFn = (
 ) => TemplateFnReturnType;
 
 export type FooterActionFn = (
-  props: unknown,
+  props: never,
   helpers: TemplateHelpers
 ) => TemplateFnReturnType;
 
@@ -114,7 +114,7 @@ function createTemplateFunction<
   R = TemplateFnReturnType,
 >(
   original: ((props: P, helpers: TemplateHelpers) => R) | undefined
-): ((props: P) => JSX.Element) | undefined {
+): ((props: P) => JSX.Element | null) | undefined {
   if (!original) return undefined;
 
   return (props: P) => {
@@ -157,6 +157,8 @@ export function createDocSearch<TComponentProps, TInputProps = TComponentProps>(
     const ref = createRef<DocSearchRef>();
     let isReady = false;
 
+    const FooterAction = createTemplateFunction(footerAction);
+
     const props: TComponentProps = {
       ...rest,
       ref,
@@ -165,7 +167,9 @@ export function createDocSearch<TComponentProps, TInputProps = TComponentProps>(
         createTemplateFunction<ResultsFooterComponentProps>(
           resultsFooterComponent
         ),
-      footerAction: createTemplateFunction(footerAction),
+      footerAction: FooterAction
+        ? createElement(FooterAction, null)
+        : undefined,
       transformSearchClient: (searchClient: unknown): unknown => {
         if (
           typeof searchClient === 'object' &&
--- a/packages/docsearch-js/src/createDocSearch.tsx
+++ b/packages/docsearch-js/src/createDocSearch.tsx
@@ -1,4 +1,8 @@
 import type { DocSearchRef, InitialAskAiMessage } from '@docsearch/core';
+import type {
+  ResultsFooterComponentProps,
+  HitComponentProps,
+} from '@docsearch/react';
 import htm from 'htm';
 import type { ComponentType, JSX, Attributes } from 'preact';
 import {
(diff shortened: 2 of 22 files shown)


LorrisSaintGenez (changes requested):
Superseded by the inline review below.

packages/website/docs/packages/react/api-reference.mdx
```diff
@@ -95,12 +95,12 @@ Transforms hits before DocSearch groups and renders them. Defaults to the identi
...
 ### `hitComponent`
 
-> `type: React.ComponentType<HitProps>` | **optional**
+> `type: ({ hit, children }) => JSX.Element` | **optional**
 
 Renders one result link. The default component renders the standard result content.
 
 ```tsx title="Hit.tsx"
-function Hit({ hit, children }: HitProps): JSX.Element {
+function Hit({ hit, children }): JSX.Element {
```
LorrisSaintGenez: **Must fix.** Dropping the annotation makes this an implicit-any tsx sample. This PR just exported `HitComponentProps`, use it here instead of deleting the type.

packages/website/docs/packages/js/api-reference.mdx
```diff
@@ -290,16 +296,26 @@ Set `translations.modal.resultsScreen.resultBadgeLabelText` to describe the badg
...
+  resultsFooterComponent({ state }, { html }) {
+    return html`<p>No results found for query: ${state.query}</p>`;
+  },
+  footerAction(_, { html }) {
+    return html`<a href="https://algolia.com">Our other project</a>`;
+  },
+});
 ```
 
-String returns render as text, not HTML.
```
LorrisSaintGenez, line 301: **Must fix.** Don't delete this. `createDocSearch.tsx:125` still wraps string returns in a `<span>`, so the caveat is still true and still the thing integrators get wrong.

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -69,38 +78,18 @@ export interface DocSearchProps {
...
-      html: (template: TemplateStringsArray, ...values: any[]) => any;
-    }
+    props: ResultsFooterComponentProps
   ) => JSX.Element | null;
+  /**
+   * A custom action that can be rendered in the Modal's footer before the
+   * Algolia logo. The component will be rendered as a child of `<div
+   * className="DocSearch-Footer-Action" />`.
+   */
+  footerAction?: () => JSX.Element | null;
```
LorrisSaintGenez: **Decide.** A zero-arg render function, rendered as a component at `Footer.tsx:111`. For React consumers `React.ReactNode` is simpler and avoids remount-on-identity-change with inline arrows, which is exactly what the docs example and `examples/demo-react/src/examples/basic-askai.tsx:53` pass. The function shape only exists to mirror the JS wrapper.

packages/docsearch-react/src/DocSearch.tsx
```diff
@@ -69,38 +78,18 @@ export interface DocSearchProps {
...
-  /**
-   * Custom component rendered at the bottom of the results panel. Supports
-   * template patterns:
-   *
-   * - HTML strings with html helper: (props, { html }) => html`<div>...</div>`
-   * - JSX templates: (props) => <div>...</div>
-   * - Function-based templates: (props) => string | JSX.Element | Function.
-   */
+  /** Custom component to render an individual hit. */
+  hitComponent?: (props: HitComponentProps) => JSX.Element;
```
LorrisSaintGenez, line 82: **Decide.** Dropping `helpers?` is the right call, React never passed it, but it breaks types. TS rejects a 2-arg callback against a 1-arg target. In-repo callers are clean, the Docusaurus adapter passes 1 arg (`adapters/docusaurus-theme-search-algolia/src/theme/SearchBar/index.tsx:190`). Call it out in the changeset.

packages/docsearch-js/src/createDocSearch.tsx
```diff
@@ -25,10 +29,64 @@ export interface DocSearchCallbacks {
   interceptAskAiEvent?: (initialMessage: InitialAskAiMessage) => boolean | void;
 }
 
+export type TemplateHelpers = Record<string, unknown> & { html: typeof html };
```
LorrisSaintGenez: **Decide.** The `Record<string, unknown>` half makes a typo'd destructure resolve to `unknown` instead of failing. Narrow to `{ html: typeof html }` while this PR is already consolidating the template types.

packages/docsearch-js/src/createDocSearch.tsx
```diff
@@ -25,10 +29,64 @@ export interface DocSearchCallbacks {
...
+  helpers: TemplateHelpers
+) => TemplateFnReturnType;
+
+export type ResultsFooterComponentFn = (
+  props: ResultsFooterComponentProps,
+  helpers: TemplateHelpers
+) => TemplateFnReturnType;
+
+export type FooterActionFn = (
+  props: unknown,
```
LorrisSaintGenez: **Decide.** `props: unknown` forces `footerAction(_, { html })` in every JS integration, which the docs example already shows. Either drop the param or type it as an empty object you can extend later.

packages/docsearch-js/src/createDocSearch.tsx
```diff
@@ -51,13 +109,11 @@ const html = htm.bind(createElement) as unknown as (
...
-
 function createTemplateFunction<
-  P extends Record<string, unknown>,
-  R = JSX.Element | string | (() => JSX.Element),
+  P = Record<string, unknown>,
+  R = TemplateFnReturnType,
 >(
-  original: ((props: P, helpers?: TemplateHelpers) => R) | undefined
+  original: ((props: P, helpers: TemplateHelpers) => R) | undefined
 ): ((props: P) => JSX.Element) | undefined {
```
LorrisSaintGenez: **Decide.** This return type lies. `R` is now `TemplateFnReturnType`, which includes `null`, and line 127 launders it through `return output as JSX.Element`. Return `JSX.Element | null` and widen the React prop types to match.

packages/docsearch-react/src/Footer.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { type JSX } from 'react';
 
 import { AlgoliaLogo } from './AlgoliaLogo';
+import type { DocSearchProps } from './DocSearch';
```
LorrisSaintGenez: **Nit.** This closes a type cycle DocSearch → DocSearchModal → Footer → DocSearch. `src/types/` already owns and re-exports the package's types, `HitComponentProps` / `ResultsFooterComponentProps` / the footer action type belong there.

packages/docsearch-react/src/Footer.tsx
```diff
@@ -105,6 +105,16 @@ export function Footer({
           </span>
         </li>
       </ul>
+      <div className="DocSearch-Footer-Actions">
+        {FooterAction && (
```
LorrisSaintGenez: **Nit.** This tests the function, not its output, so a `footerAction` returning null still renders an empty `<div class="DocSearch-Footer-Action">`. Testing scenario 2 ("nothing extra renders") isn't actually met. No visual impact, I measured it, but it's a DOM artifact anyone styling that selector will hit.

packages/docsearch-react/src/DocSearchModal.tsx
```diff
@@ -304,7 +304,12 @@ export function DocSearchModal({
           }}
         />
       }
-      footer={<Footer translations={footerTranslations} />}
+      footer={
+        <Footer
+          translations={footerTranslations}
+          footerAction={props.footerAction}
```
LorrisSaintGenez: **Nit.** Only modal prop read off the rest bag. Every sibling is destructured at line 52.

packages/docsearch-js/src/__tests__/createDocSearch.test.tsx
```diff
@@ -0,0 +1,108 @@
+import { createElement, type JSX } from 'preact';
```
LorrisSaintGenez, line 1: **Nit.** Every other package puts tests in `src/__tests__/`. Move it to `packages/docsearch-js/src/__tests__/createDocSearch.test.tsx`.

LorrisSaintGenez (changes requested):
One finding has no diff line to hang off:

**Must fix** `packages/website/docs/packages/modal/api.mdx:122` and `:128` still document `(props, helpers?) => …`, the signature this PR removes, and the page never mentions `footerAction`. `@docsearch/modal` inherits both from `ReactDocSearchModalProps`.

_Review posted by Claude Code_ 🤖

packages/docsearch-js/src/createDocSearch.tsx
8bittitan: Dropping it creates more changes to how we build the templates in `@docsearch/js`. Typing as `{}` not only breaks out lint rules, but it's also sort of a lie. I think `unknown` makes the most sense here tbh.

packages/website/docs/packages/js/api-reference.mdx
8bittitan: It being wrapping in a `<span>` makes the statement a lie then imo. Also it reads really odd and is out of place imo.

packages/website/docs/packages/react/api-reference.mdx
8bittitan: Eh fair enough. Figured it was fine for just a small inline example 😅

packages/docsearch-react/src/DocSearch.tsx
8bittitan: `v5` is a breaking change major, I think it's fine.

packages/docsearch-js/src/createDocSearch.tsx
8bittitan: Actually, `never` is the correct type for this.

packages/docsearch-react/src/DocSearchModal.tsx
8bittitan: Fixed 🫡

packages/docsearch-js/src/__tests__/createDocSearch.test.tsx
8bittitan: Fixed 🫡

packages/docsearch-react/src/Footer.tsx
8bittitan: Fixed 🫡

packages/docsearch-react/src/Footer.tsx
8bittitan: Fixed by using `React.ReactNode` 🫡

packages/docsearch-js/src/createDocSearch.tsx
8bittitan: Fixed 🫡

packages/docsearch-js/src/createDocSearch.tsx
8bittitan: Fixed 🫡

packages/docsearch-js/src/createDocSearch.tsx
8bittitan: Fixed with `never` 🫡

packages/docsearch-react/src/DocSearch.tsx
8bittitan: Fixed 🫡

packages/website/docs/packages/react/api-reference.mdx
8bittitan: Fixed 🫡

LorrisSaintGenez (approved):

PR #2884, opened 2026-05-18
feat(askai): Split Ask AI modal into own component

8bittitan:
## Summary

Splits the Ask AI experience out of `DocSearchModal` into its own dedicated `DocSearchAskAiModal` component, and refactors the React package so that keyword search and Ask AI share a common modal shell, search box, and start screen primitives.

This is primarily an internal refactor with a new public entrypoint. The default `DocSearchModal` behavior is unchanged for consumers who don't opt in to the standalone Ask AI modal.

## Motivation

`DocSearchModal.tsx` had grown to ~900 lines and mixed two distinct flows (keyword search and Ask AI). That made it hard to:

- ship Ask AI independently (bundle size + lazy loading),
- reason about lifecycle/state for each flow,
- reuse the modal shell, search box, and start screen layout.

### New public component / entrypoint

- `packages/docsearch-react/src/DocSearchAskAiModal.tsx` — standalone Ask AI modal (+505).
- `packages/docsearch-modal/src/DocSearchAskAiModal.tsx` — modal-package wrapper.
- New `askai` subpath entry in `@docsearch/react` and `@docsearch/modal` (`askai.js`, `tsdown.config.ts`, `package.json` exports).
- New `DocSearchAI` component (`packages/docsearch-react/src/DocSearchAI.tsx`).

### Slimmed-down `DocSearchModal`

- `DocSearchModal.tsx`: **-809 / +96**. Ask AI-specific logic moved out; remaining file orchestrates the keyword flow.

### Shared modal primitives (`src/modal/`)

Hooks and helpers reused by both modals:

- `useDocSearchModalEffects`, `useDocSearchInsights`, `useModalEnvironment`, `useModalRefs`, `useInitialModalQuery`, `useRefreshOnInitialQuery`, `useSaveRecentSearch`, `useStoredDocSearches`
- `createDocSearchSources`, `createAskAiSources`, `normalizeDocSearchIndexes`
- `DocSearchModalShell` — shared modal scaffolding/layout.

### Shared UI primitives (`src/components/ui/`)

- `SearchBoxForm` — shared form behavior for both search boxes.
- `RecentConversationsResults` — Ask AI-scoped recent results component.
- Split `AskAiSearchBox` (-145/+77) and new `KeywordSearchBox`.
- New `AskAiStartScreen` and `KeywordStartScreen`; `ScreenState` simplified (-65/+9).
- `StoredSearchesSections` simplified (-44/+12).

### Tests

New focused tests covering the split:

[...]

PR #2980, opened 2026-08-07
fix: Add missing export for search only

8bittitan:
## What

Fixes missing published export for keyword only `<DocSearch />` that was missed in [#2884](https://github.com/algolia/docsearch/pull/2884).


LorrisSaintGenez (approved):
good catch 👀

PR #3052, opened 2026-09-17
fix(react): prevent Safari 14 crash from lookbehind regex

### LINKED ISSUE #3051: [sidepanel-js] ESM bundle crashes on Safari 14 because RegExp lookbehind detection is pre-evaluated
## Description

Importing `@docsearch/sidepanel-js` causes the entire application to fail during module evaluation on Safari 14.1.2.

Safari reports:

```text
SyntaxError: Invalid regular expression: invalid group specifier name

RegExp
getRegex
(module code)
moduleEvaluation

### PR DESCRIPTION
**Pull request by 8bittitan**
## Why

Fixes #3051 

Importing `@docsearch/sidepanel-js` (and other bundles that include Marked)
crashed on Safari 14.1.2 during module evaluation:

```
SyntaxError: Invalid regular expression: invalid group specifier name
```

tsdown/rolldown was optimizing away Marked's runtime lookbehind feature
detection. The build tool treated the `new RegExp('(?<=1)(?<!1)')` probe
as dead code and inlined it to `true`, so Marked always picked a regex
pattern that relies on lookbehind. Safari 14 doesn't support lookbehind,
so constructing that pattern threw and broke module evaluation entirely.

This is a known upstream bug in rolldown's minifier (fixed in 1.2.1). We
were pinned to tsdown 0.22.7 (rolldown 1.1.5), which predates the fix.

Separately, `extractLinksFromMessage` in the Ask AI link extractor used
its own unconditional negative lookbehind regex, which would throw the
same way on Safari 14 once an AI response contains a URL.

## What changed

- Upgraded tsdown to 0.23.0 (pulls in the rolldown fix) via a Bun catalog
  entry, so all packages share one pinned version.
- Removed the lookbehind from `extractLinksFromMessage` by masking
  matched markdown links before scanning for bare URLs, instead of using
  a lookbehind to skip them.
- Adjusted the `docsearch-modal` UMD bundle size budget to reflect the
  new output size after the toolchain upgrade.

## Testing

- Load `@docsearch/sidepanel-js` in Safari 14.1.2 (or an equivalent old
  WebKit build) and confirm the import no longer throws.
- Open the Ask AI sidepanel, ask a question that returns a response with
  both a markdown link and an adjacent bare URL, and confirm both links
  are extracted correctly with no duplicates.
- Run the existing unit tests for `extractLinksFromMessage` to confirm
  markdown and bare URL extraction still behaves as expected.

### CODE REVIEW

**Review by pipeline1987 — APPROVED** [review]

File: packages/docsearch-core/package.json [hunk 1]
```diff
@@ -39,7 +39,7 @@
     "@testing-library/jest-dom": "6.6.3",
     "@testing-library/react": "16.2.0",
     "preact": "11.0.0-beta.0",
-    "tsdown": "0.22.7",
+    "tsdown": "catalog:dev",
```
pipeline1987, line 42: catalogs 🔥

pipeline1987 (approved):

LorrisSaintGenez (approved):

